PR #2, opened 2026-04-08
Add Sonnet 7-day usage bar to the OAuth quota view

### LINKED ISSUE #1: Support Sonnet 7-day usage in the quota view
## Summary
The TUI currently shows the existing OAuth/Max quota bars, but it does not surface the Sonnet-specific 7-day usage bucket when available.

## Proposed change
- Keep the current session and weekly bars
- Add a Sonnet 7-day bar to the OAuth/Max account view when the upstream OAuth usage payload exposes `seven_day_sonnet`
- Keep this as a view-only addition without changing account switching behavior

## Why
Claude subscription usage can expose a separate Sonnet weekly bucket, and showing it in TeamClaude makes the dashboard more accurate for Sonnet-heavy usage.

guilhermesilveira:
## Summary
- add OAuth usage parsing for `seven_day_sonnet`
- display an extra Sonnet 7-day bar in the TUI and status output when available
- keep switching behavior unchanged and use the new bucket for view only

## Why
TeamClaude already shows the existing OAuth quota bars, but Claude subscription usage can expose a separate Sonnet 7-day bucket. Surfacing that bucket makes the dashboard more accurate for Sonnet-heavy usage.

Closes #1

## Validation
- `node --check src/index.js`
- `node --check src/tui.js`
- `node --check src/oauth.js`
- `node --check src/account-manager.js`

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Sits transparently between Claude Code and the Anthropic API, managing multiple 
 
 - **Automatic account rotation** — switches to the next account when session (5h) or weekly (7d) quota reaches the configured threshold (default 98%)
 - **Auto-retry on 429** — waits the `retry-after` duration and retries the same account; switches to the next on persistent errors
-- **Interactive TUI** — real-time dashboard with color-coded quota bars, reset countdowns, activity log, and keyboard controls
+- **Interactive TUI** — real-time dashboard with color-coded quota bars, reset countdowns, activity log, and keyboard controls, including Sonnet weekly usage when available
 - **OAuth token management** — automatically refreshes tokens nearing expiry and persists them to config; client token refreshes pass through untouched
 - **Hot-reload accounts** — add accounts via `import` or `login` while the server is running, press **R** to pick them up
 - **Account deduplication** — detects duplicate accounts by UUID and keeps the most recent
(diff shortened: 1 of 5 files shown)


MagicalTux:
Thanks for this, and apologies for the slow turnaround. We're going to take a different route, so closing this PR — but the Sonnet-bar idea (#1) is still on the table.

The blocker is a design constraint: **TeamClaude is passive-only — it never originates calls to the Anthropic API, it only observes the traffic Claude Code sends through the proxy.** This PR adds a 60s `fetchUsage` poll of `/api/oauth/usage`, which crosses that line. It also refreshes the rotation-driving `unified5h`/`unified7d` quota from that endpoint, which changes switching behavior, whereas #1 asked for a view-only addition.

The Sonnet 7-day bucket isn't present in the `anthropic-ratelimit-unified-*` response headers we already read passively — it only lives in the `/api/oauth/usage` JSON. So the path forward is to capture `seven_day_sonnet` **passively** if/when it appears in observed traffic (a Sonnet rate-limit header, or a usage response that Claude Code itself routes through the proxy), and render the bar only when present — a true no-op otherwise.

The normalization helpers here (`normalizeUsageBucket`, the bucket field handling) are useful reference for that passive parsing. Much appreciated.

PR #4, opened 2026-05-03
feat: persist quota state across restarts and add TUI account reorder

rlawjdghksdlqslek:
## Summary                                                                                  

  Two related quality-of-life improvements for TeamClaude's account management:               
   
  **1. Persist quota state across server restarts**                                           
                                                                                            
  Previously, quota counters (5h/7d unified usage, token limits, reset timestamps)            
  were reset to zero every time the server restarted. This made it impossible to            
  track remaining quota accurately after a restart.                                           
                                                                                              
  - Adds `exportQuota()` to `AccountManager` to serialize quota fields safe for config storage
  - Saves quota state to `savedQuota` in the config file every 60 seconds via a periodic      
  interval                                                                                    
  - Restores saved quota on startup when loading accounts                                   
  - Flushes quota state to disk on graceful shutdown (SIGINT, SIGTERM, and TUI quit)          
                                                                                              
  **2. TUI account reorder with `o` key**                                                     
                                                                                              
  Adds an interactive reorder mode to the TUI so users can reorganize the account             
  list without manually editing the config file.
                                                                                              
  - Press `o` in normal mode to enter reorder mode                                          
  - Navigate with `↑`/`↓` (or `j`/`k`), press `Enter` to grab an account
  - Move the grabbed account up/down, press `Enter` to confirm or `Esc` to cancel             
  - Saves the new order to config on confirm                                                  
  - Updates `currentIndex` correctly when accounts are moved around                           
                                                                                              
  ## Files Changed                                                                          
                                                                                              
  - `src/account-manager.js` — `exportQuota()`, `moveAccount()`, quota restore on init        
  - `src/index.js` — periodic quota save, flush on shutdown
  - `src/tui.js` — `reorder` mode, `_keyReorder()`, `_doSaveReorder()`, footer hint update    
                                                                                              
  ## Test Plan
                                                                                              
  - [ ] Start server, make some requests to accumulate quota usage, restart — verify quota    
  values are preserved in TUI
  - [ ] In TUI, press `o`, reorder accounts, confirm — verify new order persists after restart
  - [ ] In TUI, press `o`, reorder, press `Esc` — verify order is reverted                    
  - [ ] Verify `currentIndex` stays on the correct account after reordering

### DIFF
--- a/src/account-manager.js
+++ b/src/account-manager.js
@@ -348,6 +348,24 @@ export class AccountManager {
   }
 
   /**
+   * Move an account from one index to another, updating currentIndex accordingly.
+   */
+  moveAccount(from, to) {
+    if (from === to || from < 0 || to < 0 ||
+        from >= this.accounts.length || to >= this.accounts.length) return;
+    const [account] = this.accounts.splice(from, 1);
+    this.accounts.splice(to, 0, account);
+    this.accounts.forEach((a, i) => a.index = i);
+    if (this.currentIndex === from) {
+      this.currentIndex = to;
+    } else if (from < to) {
+      if (this.currentIndex > from && this.currentIndex <= to) this.currentIndex--;
+    } else {
+      if (this.currentIndex >= to && this.currentIndex < from) this.currentIndex++;
+    }
+  }
+
+  /**
    * Return a status summary of all accounts (safe to expose, no credentials).
    */
   getStatus() {
--- a/src/account-manager.js
+++ b/src/account-manager.js
@@ -28,7 +28,7 @@ export class AccountManager {
       refreshToken: acct.refreshToken || null,
       expiresAt: acct.expiresAt || null,
       status: 'active',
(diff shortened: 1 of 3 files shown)


MagicalTux:
Thanks for this — both ideas were good. Superseding it with two focused changes rather than merging directly, since the branch predates a lot of churn (org-aware accounts, rotation priority) and is conflicting.

- **Quota persistence** → reimplemented in #30. Same goal, but the runtime quota is written to a separate `teamclaude.state.json` sibling rather than a `savedQuota` blob in `config.json`, so the hand-editable config stays clean and isn't rewritten every 60s. State is keyed by account identity and stale windows are discarded on restore. You're credited in the PR.
- **TUI account reorder** → dropped, because rotation order is now controlled by the `priority` command/field (#28). The reorder save path here also used the pre-org-aware identity matching, which would mis-handle one email across multiple orgs.

Really appreciate the contribution.

PR #8, opened 2026-05-06
Show quota reset times in CLI status output

brendandebeasi:
Extracts the `formatReset` helper into `src/index.js` and updates the `teamclaude status` command to display time-to-reset (e.g., "resets in 2h30m") next to the usage percentages for both unified and standard quotas.

Matches the behavior already present in the TUI dashboard.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -384,13 +384,17 @@ async function statusCommand() {
       console.log(`    Status:   ${acct.status}`);
 
       if (q.unified5h != null || q.unified7d != null) {
-        const ses = q.unified5h != null ? (q.unified5h * 100).toFixed(1) + '%' : '-';
-        const wk = q.unified7d != null ? (q.unified7d * 100).toFixed(1) + '%' : '-';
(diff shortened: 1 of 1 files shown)


brendandebeasi:
Note on behavior: When testing this on an active gateway, I noticed that inactive accounts might show a usage percentage without a corresponding reset time. This is expected behavior: if the reset time has already passed, TeamClaude clears the timestamp, but the usage percentage remains "frozen" at its last known value until that account is actually used again to fetch fresh headers from Anthropic.

brendandebeasi:
Here is an example of what the new output looks like when tested on a live gateway:

```text
Active account: E1
Switch at:      98% usage

E1 (oauth) *
    Status:   active
    Session:  21.0% (resets in 4h14m)    Weekly: 38.0% (resets in 1d7h)
    Total:    1069291 tokens, 319 requests

E2 (oauth)
    Status:   active
    Session:  22.0% (resets in 3h24m)    Weekly: 98.0%
    Total:    865559 tokens, 2364 requests
```

PR #10, opened 2026-05-07
Add `shim` command for transparent claude→proxy routing

teocns:
## Summary

Adds a `teamclaude shim` subcommand that drops a tiny bash wrapper at `$XDG_DATA_HOME/teamclaude-shim/claude` and wires it onto every shell's `PATH` via the **rustup-style env-file pattern**. Plain `claude` invocations then probe the proxy port locally — applying `teamclaude env` and exec'ing the real `claude` if the proxy is up, or exec'ing the real `claude` directly if it isn't.

The shim lives in its own directory, separate from where Claude Code's auto-updater rewrites its binary, so it survives `claude` self-updates indefinitely (same trick `rbenv`, `asdf`, and `mise` use to survive language-version updates).

## Why

`teamclaude run` covers ad-hoc invocations, but anyone who types `claude` directly — from a shell, an editor, a script, a hotkey — bypasses the proxy. A naive shadow-symlink in the same dir as `claude` works for a day, then gets clobbered the next time Claude Code self-updates. The PATH-prepended shim dir avoids that collision class entirely.

## Surface

```
teamclaude shim install     [--no-rc] [--shim-dir PATH]
teamclaude shim uninstall
teamclaude shim status      # also: bare `teamclaude shim`
```

## Shell coverage

Modeled on `~/.cargo/env` (the rustup install pattern). Three files written under the shim dir:

- `claude` — the wrapper (executable)
- `env` — POSIX sh loader, idempotent at source-time via `case ":${PATH}:" in *:"...":*) ;;`
- `env.fish` — fish loader using `fish_add_path`-style logic

Then a single `. "<shim-dir>/env"` line is appended to each rc:

| Shell | rc file(s) wired |
| --- | --- |
| bash | `~/.bashrc` AND `~/.bash_profile` (covers macOS Terminal's login-shell precedence) |
| zsh | `~/.zshrc` |
| POSIX sh | `~/.profile` |
| fish | `~/.config/fish/conf.d/teamclaude-shim.fish` (auto-loaded; no rc edit) |

The loaders are idempotent at source time (PATH dedup via case statement / `not contains`) so re-sourcing is safe across shell reloads. Uninstall surgically strips the comment + source line from each rc and removes all written files.

## Implementation notes

- **Zero new dependencies.** New module `src/shim.js` uses only `node:fs`, `node:path`, `node:os`. The bash wrapper and shell loaders are embedded as template literals and written at install time — no shell scripts in the repo, consistent with the existing pure-Node layout.
- **Dispatch.** Adds one `case 'shim'` to the switch in `src/index.js`. New `shimCommand()` parses `install` / `uninstall` / `status` and forwards flags. Help text and README updated to match existing CLI vocabulary.
- **Smart rc selection.** Existing rc files are always wired; non-existent ones are only created if they match the user's current `$SHELL` (avoids creating `.bashrc` on a zsh-only machine).
- **`$HOME` portability.** When the shim dir lives under `$HOME`, env files reference it as `$HOME/...` (rustup convention) for portability across machines on shared/network homes.
- **Real-binary resolution.** Wrapper walks `PATH`, skipping any entry whose realpath matches itself — works regardless of where Claude Code installs (npm, homebrew, native installer).

## Tested

- `node --check` on both modified files; existing `eslint` clean (no new warnings).
- End-to-end install → idempotent re-install → uninstall against a sandboxed `$HOME` populated with pre-existing rc content — original content preserved verbatim, no duplicate source lines on re-install, full cleanup on uninstall.
- Functional shell sourcing verified: sourcing `env` from bash exports `PATH` correctly; sourcing it twice does not duplicate the entry (case-statement idempotence works).
- Live execution against a running proxy: shim probes `:3456`, eval's `teamclaude env`, exec's real `claude --version` successfully through the proxy.
- Live execution with proxy down: shim skips env injection, exec's real `claude --version` directly with a clean env.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,12 +122,53 @@ eval $(teamclaude env)
 claude
 ```
 
+### Transparent shim
+
+`teamclaude run` works for one-off invocations, but if you want plain `claude` to route through the proxy automatically — without prefixing every call — install the shim:
+
+```bash
+teamclaude shim install
+```
+
+This drops a small bash wrapper at `$XDG_DATA_HOME/teamclaude-shim/claude`, alongside `env` (sh / bash / zsh) and `env.fish` (fish) loaders. Each detected shell rc gets a single one-line directive sourcing the loader — same pattern rustup uses with `~/.cargo/env`. From then on, every `claude` invocation:
+
+1. Probes the proxy port locally.
+2. **Up** → applies `teamclaude env` and execs the real `claude`.
+3. **Down** → execs the real `claude` directly.
+
+The shim lives in its own directory, separate from where Claude Code's auto-updater writes its binary. So `claude` updates can come and go without disturbing the shim — same trick `rbenv`, `asdf`, and `mise` use to survive language-version updates.
+
+```bash
+teamclaude shim status     # Show install state and which rc files are wired up
+teamclaude shim uninstall  # Revert (removes shim files + cleans rc edits)
+```
+
+Shells covered:
+
+- **bash** — `~/.bashrc` and `~/.bash_profile` (handles macOS Terminal's login-shell precedence)
+- **zsh** — `~/.zshrc`
+- **POSIX sh** — `~/.profile` (login-shell baseline; helps display managers, etc.)
+- **fish** — `~/.config/fish/conf.d/teamclaude-shim.fish` (auto-loaded; no rc edit)
+
+The sourced loaders are idempotent at source time (they check whether the shim dir is already on `PATH`) so reload-after-reload is safe.
+
+Flags:
+
+- `--no-rc` — skip rc edits; print the source lines for manual install.
+- `--shim-dir PATH` — override the install directory (default `$XDG_DATA_HOME/teamclaude-shim`).
+
+Shim runtime env vars:
+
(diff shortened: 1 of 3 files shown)


MagicalTux:
Thanks for this — the analysis (PATH-prepended dir to survive Claude Code's auto-updater, rustup-style env files, multi-shell rc handling) is excellent and was the reference for the direction we took.

We've gone with a lighter approach in #33 for the common "route `claude` when I type it" case:
- `teamclaude run` now probes the proxy and falls back to launching `claude` directly when it's down (so the routing logic lives in one place).
- `teamclaude alias` prints `alias claude='teamclaude run --'` (or `--install` into the shell rc).

The trade-off vs this PR: an interactive-shell alias doesn't intercept `claude` spawned by editors/scripts the way the PATH shim does — but it's ~110 lines vs ~340, shadows no binary, edits one reversible line per rc, and is a no-op for anyone who doesn't opt in. For users who specifically need editor/script interception, the PATH-shim approach here is still the right tool, and we may revisit it.

Really appreciate the thorough write-up and implementation.

PR #19, opened 2026-06-01
feat(ui): responsive dashboard with quota reset times

jphein:
## Summary
Brings the web dashboard (`/teamclaude/ui`) to a polished, responsive state and surfaces quota **reset times** alongside utilization.

- **Responsive dashboard** — fluid card grid with container queries; layout adapts down to narrow cards (`08c80a7`).
- **Reset-time display** — each 5h/7d quota bar now shows when its window rolls over (`↺ HH:MM` within a day, weekday + time beyond), with the full timestamp on hover. Reads `unified5hReset`/`unified7dReset` (epoch-ms from Anthropic rate-limit headers) already exposed by `getStatus()`, with a fallback to legacy `resetsAt` (`444281e`).
- **Fast shutdown + live HTML reload** — server reads `web/index.html` from disk per request (`Cache-Control: no-store`), so dashboard edits are live without a restart (`08c80a7`).

## Pre-Landing Review
No issues found. Verified the reset feature is wired end-to-end: backend stores reset timestamps as epoch-ms (`account-manager.js:173-174`), `getStatus()` forwards the full quota object via spread (`account-manager.js:340`), and the frontend `fmtReset(ms)` consumes ms directly and guards null/past values. Units consistent on both sides. Secret scan clean.

## Test plan
- [ ] No automated test suite in this project (`package.json` defines only `start`).
- [ ] Manual: open `http://localhost:3456/ui` against a running proxy with active accounts and confirm reset labels render next to each bar and resize cleanly on a narrow window.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,10 +128,22 @@ claude
 teamclaude accounts          # List accounts with subscription tier and token status
 teamclaude accounts -v       # Also show token expiry times
 teamclaude status            # Show live proxy status (requires running server)
+teamclaude switch <name>     # Manually pin the active account
+teamclaude threshold <val>   # Set rotation threshold ("85", "85%", or "0.85")
 teamclaude remove <name>     # Remove an account
 teamclaude api <path>        # Call an API endpoint with account credentials
 teamclaude help              # Show all commands
 ```
+
+### Web Dashboard
+
+When the proxy is running headless (e.g. as a systemd service) the TUI isn't available. The proxy serves a single-page dashboard on the same port:
+
+```
+http://localhost:3456/ui
+```
+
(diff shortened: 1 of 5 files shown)


MagicalTux:
I'm not sure the web dashboard adds much compared to the TUI, do you have a rationale for adding it?

jphein:
Probably a silly reason, but I run team Claude as a service for my lan. When made into a systemd service I can no longer use the tui. And also I can view the website via other hosts using the proxy on the lan.

wolph:
Seems like a very useful addition to me. I was actually considering building something like this myself.

In my case it's because teamclaude runs as a service so I don't have any access to the TUI interface and `teamclaude status` shows very few details (when was it last updated? is probing enabled? how much time till the next probe?)

MagicalTux:
Closing as stale: this branch conflicts with master and predates substantial changes (e.g. it assumes no automated test suite exists, which is no longer true). Please rebase and open a fresh PR if the responsive dashboard work is still desired.

PR #22, opened 2026-06-09
fix: strip fast mode so it can't poison the account pool

jphein:
## Summary
Toggling `/fast` in Claude Code made the proxy appear dead — every account reported "rate limited."

**Root cause.** `/fast` sends `speed:"fast"` + a `fast-mode-*` beta header, routing the request through a priority-tier pool. On a subscription seat that pool bills as *usage credits* and is **not** covered by subscription rate limits, so a seat without credits gets a **429 on every fast request, regardless of account**. The proxy treated each 429 as quota exhaustion: it rotated to the next account and rate-limited the previous one for the upstream `retry-after`. A single `/fast` request thus cascaded through and penalized **all** accounts for ~60s.

**Fix.**
- `stripFastMode()` removes `speed:"fast"` and the `fast-mode-*` beta token before forwarding, re-syncing `content-length`. The request runs as standard Opus and never hits the priority pool — `/fast` becomes a safe no-op on subscription seats.
- The 429 branch no longer discards the upstream body — it logs the reason, so future non-quota 429s are diagnosable from journald instead of being a silent blind spot.

## Pre-Landing Review
No issues found. JSON parsing and the 429 body read are both `try/catch`-guarded; no secrets; no dead code; `content-length` correctly re-synced after the body shrinks.

## Test plan
- [x] 9/9 tests pass (`node --test`), incl. new regression tests: fast-mode body is stripped + beta token removed while unrelated betas survive; non-fast requests forwarded byte-for-byte unchanged.
- [x] Deployed to the live systemd user service (npm-link), restarted cleanly with all 4 accounts loaded.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,10 +128,22 @@ claude
 teamclaude accounts          # List accounts with subscription tier and token status
 teamclaude accounts -v       # Also show token expiry times
 teamclaude status            # Show live proxy status (requires running server)
+teamclaude switch <name>     # Manually pin the active account
+teamclaude threshold <val>   # Set rotation threshold ("85", "85%", or "0.85")
 teamclaude remove <name>     # Remove an account
 teamclaude api <path>        # Call an API endpoint with account credentials
 teamclaude help              # Show all commands
 ```
+
+### Web Dashboard
+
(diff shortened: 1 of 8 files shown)


jphein:
Closing — this fix is for my own fork only, not intended for upstream.

PR #37, opened 2026-06-18
Opt-in MITM forward-proxy mode (intercept hardcoded api.anthropic.com)

MagicalTux:
Adds an **opt-in, off-by-default** HTTPS forward-proxy mode so teamclaude can intercept Claude Code traffic that uses a **hardcoded** `api.anthropic.com` URL (e.g. the **Claude Design MCP**), which ignores `ANTHROPIC_BASE_URL` and so bypasses the normal reverse proxy. With a placeholder token in the local claude, those requests 401; routed through us, the real account token is injected.

## How it works
- `teamclaude run` (when `useMitmProxy: true`) launches claude with `HTTPS_PROXY`/`HTTP_PROXY` → teamclaude, and `NODE_EXTRA_CA_CERTS` → a locally-generated CA.
- claude sends `CONNECT api.anthropic.com:443`; teamclaude TLS-terminates with a leaf cert (trusted via the CA), and the decrypted request flows through the **existing `forwardRequest`** (token injection + forward to the real upstream).
- **Everything that isn't the upstream host is blind-tunneled untouched.**
- The server **always** accepts both base-URL and proxy/CONNECT clients, so multiple claude instances with different settings can share one server.

## No new dependencies
Certificates are generated by a small **pure-JS X.509** implementation (`src/x509.js`, ASN.1 DER + `node:crypto`) — **no openssl, no npm deps**. Verified by parse + chain-verify + a real TLS handshake in tests.

## Built-in test endpoint (curl-friendly)
The proxy always intercepts `www.example.org` and answers a canned JSON, so the proxy + CA can be verified with **no credentials**:
```bash
curl --proxy http://localhost:3456 --cacert ~/.config/teamclaude-ca.pem https://www.example.org/
# {"teamclaude":"mitm-proxy-ok",...}
```

## Trust model
- CA generated locally, stored in the config dir, trusted **only** by the launched claude process (via `NODE_EXTRA_CA_CERTS`) — never added to the system store. Leaf key `0600`; the **CA private key is never written to disk** (chain regenerated wholesale when needed).
- Upstream (teamclaude → real Anthropic) TLS verification stays **on**.
- **Off by default** (`useMitmProxy: false`) — to be tested extensively before it could ever default.

## Tests
`test/x509.test.js` (cert chain, SAN, TLS handshake trust/reject) and `test/mitm.test.js` (ensureCerts idempotency + multi-SAN, MITM round-trip via the test host with handshake authorized, blind tunnel). Full suite **63 passing**, lint clean. Also verified manually via `curl` end-to-end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Sits transparently between Claude Code and the Anthropic API, managing multiple 
 - **Enable/disable accounts** — temporarily pause an account without removing it (`teamclaude disable`/`enable`, or `d` in the TUI); re-enabling also clears a stuck error state
 - **Quota persistence** — observed quota survives restarts (saved to a sibling state file), so rotation state isn't lost on restart; stale windows are discarded automatically
 - **Optional quota probe** — off by default; when enabled, periodically refreshes idle accounts' quota from the usage endpoint (no message spend), and surfaces the Sonnet weekly bucket
+- **Optional MITM proxy mode** — off by default; routes claude via an HTTPS forward proxy with a local CA so even hardcoded `api.anthropic.com` endpoints (e.g. the Claude Design MCP) get the real token injected
 - **Request logging** — optional full request/response logging for debugging
 - **Zero dependencies** — uses only Node.js built-in modules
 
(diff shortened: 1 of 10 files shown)


MagicalTux:
Update: per review, MITM is now a **`run` flag** rather than a config option (the PR title/body above predate this — `gh pr edit` is currently blocked by a projects-classic API deprecation).

- Enable per-launch: `teamclaude run --mitm -- <claude args...>` (flag recognized only before `--`).
- Removed the `useMitmProxy` config field.
- `teamclaude alias --install --mitm` emits a `run --mitm --` alias for plain `claude`.
- Server is unchanged — it already always accepts both base-URL and CONNECT clients, so instances with and without `--mitm` share one server.

Full suite 64 passing; verified end-to-end via `curl --cacert` against the built-in `www.example.org` test endpoint.

PR #49, opened 2026-06-25
fix 1% usage being parsed into 100%

binary-person:
`https://api.anthropic.com/api/oauth/usage` should always return values between 0-100, not 0-1. I think it might have been claude overengineering the solution


MagicalTux (approved):
Verified against the live /api/oauth/usage endpoint: utilization is a 0–100 percentage (observed 2.0=2%, 9.0=9%, corroborated by limits[].percent). The previous >1 heuristic stored values ≤1% as up to 100%. Fix is correct; full suite passes locally (127/127).

PR #83, opened 2026-07-06
Pass attachment transfers (/api/oauth/files/*) through with the client's own credential (fixes silently dropped images)

iamapfelbaum:
## Problem

An image attached to a message for a teleoperated Claude Code session (claude.ai / desktop app → session behind teamclaude) is silently lost: the message arrives as plain text, with no image block and not even an `[Image #N]` placeholder.

## Cause

The attachment is uploaded to the backend under the user's claude.ai identity, and the CLI then downloads it with a separate request — `GET /api/oauth/files/<uuid>/content` — using its own OAuth token. This path is not covered by the `/v1/code/` passthrough (#58): the proxy rewrites `Authorization` to the rotated account's token, upstream responds 403, and Claude Code silently drops the image (in the CLI code this is a buried skip with no error surfaced to the user).

The same logic applies to `/api/oauth/file_upload`.

## Fix

`/api/oauth/files/` and `/api/oauth/file_upload` now go through `relayStream` with the client's own credential — exactly like the Remote Control channel. The prefixes are extracted into a shared `CLIENT_CREDENTIAL_PATHS` list.

## Verification

Verified end-to-end on a live install (Contabo VPS, 3-account rotation): before the patch, attachments never reached server-side sessions (while sessions without the proxy received them fine). After the patch, an image attached from the desktop app arrived in the server-side session as a full base64 image block, and file attachments landed in `~/.claude/uploads/<session>/`.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -140,6 +140,10 @@ export function resolveAccountPin(accountManager, token) {
   return null;
 }
 
+// Paths that must reach upstream with the client's own credential (never a
+// rotated account token): the Remote Control channel and attachment transfers.
+const CLIENT_CREDENTIAL_PATHS = ['/v1/code/', '/api/oauth/files/', '/api/oauth/file_upload'];
+
 export function createProxyRequestListener({ accountManager, upstream, logDir = null, hooks = {}, sx = null }) {
   let counter = 0;
   return async (req, res) => {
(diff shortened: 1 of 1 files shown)


iamapfelbaum:
Closing until the fix is verified end-to-end on a live install — will reopen once confirmed.

iamapfelbaum:
Verified end-to-end on a live install (3-account rotation): after routing /api/oauth/files/ through relayStream, an image attached from the desktop app arrived in the server-side session as a full base64 image block, and file attachments landed in ~/.claude/uploads/<session>/. Before the patch the same flow silently dropped every attachment.

PR #87, opened 2026-07-08
docs: explain prompt-cache behavior across account rotation

SihyeonJeon:
Documents a non-obvious interaction the README doesn't cover (grep 'cache' in README = 0 hits). Rotation preserves conversation context (client resends transcript + account_uuid rewrite), but Anthropic's prompt cache is org-scoped so the first turn after a switch is a full cache miss before the new account re-warms. Ties mitigations to existing features (98% threshold, teamclaude priority, keep-warm) and clarifies keep-warm warms the 5h session timer, not the prompt cache. Docs-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,19 @@ TLS is established **end-to-end with `api.anthropic.com` over the tunnel**, so t
 8. If all accounts are exhausted, returns 429 with the soonest reset time
 9. Client token refresh requests (`/v1/oauth/token`) are relayed to upstream untouched — the proxy and client manage their own token lifecycles independently
 
+## Prompt caching & rotation
+
+Rotation is transparent to your Claude Code session — but it is worth knowing how it interacts with Anthropic's prompt cache.
+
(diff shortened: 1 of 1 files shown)


MagicalTux:
Thanks for this — documenting how rotation interacts with the prompt cache is a genuinely useful addition, and the first two points (context is preserved via transcript resend + `account_uuid` rewrite; the cache is org-scoped so a switch costs a first-turn miss) are spot on.

Closing because the third mitigation conflates two unrelated things: **keep-warm** starts an account's **5-hour session timer**, not its prompt cache — a freshly-rotated account still takes a cache miss on its first turn regardless. The more important point is that TeamClaude is *sticky by default* (it stays on one account and prefers the soonest-expiring weekly quota), so switches — and therefore cache misses — are already rare.

I've folded the accurate version into the README under a new **"Prompt caching across rotation"** section (next to the keep-warm docs). Appreciate you flagging the gap!

PR #95, opened 2026-07-08
Multi-provider support: OpenAI/Codex alongside Claude (Phase 0: provider seam)

MagicalTux:
**Draft — in progress.** Foundational work to let teamclaude proxy **both** Claude and OpenAI/Codex from one instance, routing each request to its own account pool by host (MITM). This PR is **Phase 0**: the provider abstraction + account router, with zero behavior change. Codex login + relay + quota land in follow-up PRs.

### Design (agreed)
- **One proxy, both providers** — MITM intercepts `api.anthropic.com` → Claude pool and `chatgpt.com` → Codex pool (`ab.chatgpt.com` blind-tunneled).
- **Two `AccountManager`s behind a router** — each pool rotates independently; `AccountRouter` presents a flattened facade to TUI/persistence/prober/warmer and hands the request path the pool for the request's host.
- **Codex = single pool** metered by the account-wide `x-codex-primary/secondary-*` windows (no per-model buckets).
- **Codex auth**: real OpenAI OAuth (`auth.openai.com`, client_id `app_EMoamEEZ73f0CkXaXp7hrann`, PKCE) + `Authorization: Bearer` / `ChatGPT-Account-ID` injection. (All endpoints/headers verified against the open-source `openai/codex` repo.)

### Phase 0 checklist — ✅ complete
- [x] **0a — provider seam.** `src/providers/{index,anthropic}.js` as the single source of truth for Anthropic host/OAuth constants; replaced ~15 scattered literals; added `providerForHost`/`providerById` + tests.
- [x] **0b — behavioral hooks.** Anthropic request logic extracted out of `forwardRequest` into provider methods (`injectAuth`, `rewriteBody`, `rateLimitHeaders`, `classify429`, usage parsing), carried on `ctx.provider`. MITM `hostMode` rewrites any known provider host.
- [x] **0c — account router.** `src/account-router.js` (one `AccountManager` per provider) wired through startup + the request path (pool resolved by host, duck-typed for tests). Anthropic-only config → one pool → byte-identical.

**Behavior-preserving throughout.** 260 tests pass (baseline 243 + 17 new), lint clean. Verified at runtime: status endpoint, account selection, TUI render, and clean shutdown all work with the router.

### Next (separate PRs)
- **Phase 1** — `src/providers/openai.js`: OpenAI OAuth login, token refresh, `chatgpt.com` host registration + `ab.chatgpt.com` tunnel, Bearer/`ChatGPT-Account-ID` injection, rotate-on-429 replay. One Codex account proxying end-to-end.
- **Phase 2** — Codex quota-aware rotation from `x-codex-*` response headers; mixed-pool TUI/status.

### DIFF
--- a/src/account-router.js
+++ b/src/account-router.js
@@ -1,0 +1,139 @@
+// Runs one AccountManager per provider and routes between them. Each pool
+// (Claude, OpenAI/Codex, …) rotates independently; the request path resolves the
+// right pool by upstream host via managerFor(), while lifecycle/UI code sees a
+// flattened view across every pool.
+//
+// With a single provider configured — today's default — the router wraps exactly
+// one manager and is behaviorally a pass-through, which is what keeps the
+// provider seam a no-op until a second provider's accounts are actually added.
+
+import { AccountManager } from './account-manager.js';
+import { providerById, providerForHost, DEFAULT_PROVIDER_ID } from './providers/index.js';
+
+export class AccountRouter {
+  // `accounts` is the flat config list; each account's optional `provider` field
+  // (default 'anthropic') decides which pool it joins. `opts` mirrors the
+  // AccountManager options and is passed through to every manager; a provider
+  // that supplies its own token-refresh gets it injected here.
+  constructor(accounts, switchThreshold = 0.98, opts = {}) {
+    // Partition by provider id, preserving first-seen order so the default
+    // provider (usually the only one) stays at index 0.
+    const groups = new Map();
+    for (const acct of accounts || []) {
+      const id = acct.provider || DEFAULT_PROVIDER_ID;
+      if (!groups.has(id)) groups.set(id, []);
+      groups.get(id).push(acct);
+    }
+    if (groups.size === 0) groups.set(DEFAULT_PROVIDER_ID, []);
+
+    this.entries = [];
+    for (const [id, group] of groups) {
+      const provider = providerById(id);
+      // A provider may override token refresh (OpenAI vs Anthropic); fall back to
+      // whatever the caller passed (or the AccountManager default).
+      const managerOpts = { ...opts };
+      if (provider.refresh && managerOpts.refreshFn === undefined) managerOpts.refreshFn = provider.refresh;
+      this.entries.push({ provider, manager: new AccountManager(group, switchThreshold, managerOpts) });
+    }
+    this.defaultManager = this.entries[0].manager;
+  }
+
+  get managers() {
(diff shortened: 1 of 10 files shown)


SebastienGllmt:
I've seen quite a few projects achieve something like this by leveraging https://github.com/router-for-me/CLIProxyAPI which maybe is how the `teamclaude run` command would work with this feature

DIodide:
> I've seen quite a few projects achieve something like this by leveraging https://github.com/router-for-me/CLIProxyAPI which maybe is how the `teamclaude run` command would work with this feature

Creating an adapter using CLIProxyAPI verbatim

MagicalTux:
Closing this: the goal it existed for — pooling Codex subscriptions alongside Claude accounts — shipped in 1.1.15 via #246, by a different mechanism.

## What landed instead

#246 took the seam at the **credential and destination** rather than at the account manager. One `AccountManager`, a `provider` field on each account, and a provider partition expressed by widening the per-request exclude set — so cursors, pinning, session affinity, probing and preemption all kept working untouched, and an Anthropic-only config produces the identical selection it did before. Requests are routed by **path** (`/v1/messages` vs `/backend-api/codex/responses`).

Since then: #281 gave each provider its own cursor (a Codex request was dragging the shared `currentIndex` across and the next Anthropic request dragged it back, re-arming storm control on every alternation), scoped the partition to **subscription** accounts only — a Claude Max token is issued to Claude and a ChatGPT token to Codex, while an API key is metered capacity with no tie to a caller — and closed a hole where `TC_ACCT` bypassed selection entirely and could serve a Codex request from a Claude subscription.

## What did NOT land, and is worth keeping

Phase 0's **host-based MITM routing**. `hostMode` still rewrites only the single configured upstream host and blind-tunnels everything else; there is no `providerForHost`. The practical consequence is that Codex works through the explicit base-URL redirect (`model_providers` in `~/.codex/config.toml`) but **not through the MITM forward proxy** — a Codex CLI that only honours `HTTPS_PROXY` has no route in.

That is a real gap this PR would have covered, and it is the one piece of the design I would not want lost with the branch. If it becomes a problem, the shape is already written here: intercept `chatgpt.com`, blind-tunnel `ab.chatgpt.com`, and let `providerForHost` decide the pool the way `providerForPath` does today.

The two-`AccountManager`/`AccountRouter` structure I would not revive. #246 demonstrated the partition is expressible as an exclusion, and one manager kept every existing behaviour working without a facade in front of it.

Closing as superseded rather than abandoned.

PR #99, opened 2026-07-12
fix(routing): make account selection see the advisor model

### LINKED ISSUE #98: Investigate /advisor fable
User report about /advisor fable not being routed properly, to investigate.

MagicalTux:
Fixes #98.

## Problem

Claude Code's `/advisor` (e.g. `/advisor fable`) does not send Fable requests. The request keeps the **executor** in the top-level `model` field and nests the advisor's model inside the tools array:

```json
{
  "model": "claude-opus-4-8",
  "tools": [{ "type": "advisor_20260301", "name": "advisor", "model": "claude-fable-5" }]
}
```

TeamClaude routed only by the top-level model, so an advisor request was treated as a plain Opus request:

- `*fable*` routes and route pins never matched (the string "fable" only appears in `tools[].model`);
- an account with a spent — or missing — Fable weekly bucket could receive an advisor call it can't serve. Upstream fails the advisor sub-inference and Claude Code then **disables the advisor for the whole session** ("The advisor tool is unavailable. Do not try to use it again.");
- the advisor silently spent the Fable bucket of whichever account executor-routing happened to pick.

(The `anthropic-beta: advisor-tool-…` header was never the problem — header forwarding already passes it through.)

## Fix

- **`src/model.js`** — `AdvisorModelFinder` / `parseAdvisorModel`: a byte-exact structural scan (same discipline as `TopLevelFieldFinder`) that only reads `type`/`model` strings that are direct fields of an element of the root `tools` array, judged on element close so field order doesn't matter. Decoys in `input_schema` or conversation text never match. Gated on a cheap `Buffer.includes('advisor')` so non-advisor traffic pays one byte search.
- **`src/account-manager.js`** — `getActiveAccount(exclude, model, advisorModel)`: the account must be eligible for **both** models — the advisor's family bucket must have headroom and any route/ownership rule for the advisor model must allow the account. Route-pin precedence: the executor's pin wins (it's the bulk of the spend); the advisor's pin applies when nothing pins the executor. When **no** account satisfies both, selection degrades to executor-only routing (with a throttled log line) so the main request keeps flowing instead of synthesizing a 429.
- **`src/server.js`** — parse the advisor model once per request, carry it in `ctx` so 429 fail-over re-selection honors it too.

## Tests

257 pass (248 baseline + 9 new): parser exactness (nested/decoy fields, field order, chunked feeding), advisor-bucket skipping, advisor route pins + precedence, route exclusivity, and the executor-only degrade path. Lint clean.


ma2t:
Thanks Mark!

PR #105, opened 2026-07-15
Sanitize orphaned tool_use/tool_result blocks before forwarding

viiqswim:
## The problem

Anthropic rejects any request where a `tool_use` block isn't answered by a matching `tool_result` in the next message, with a non-retryable 400:

```
messages.N: `tool_use` ids were found without `tool_result` blocks immediately
after: toolu_XXXX. Each `tool_use` block must have a corresponding `tool_result`
block in the next message.
```

When a client compacts (summarizes) a long conversation, or an in-flight tool call gets interrupted, the slice can split that pair: a `tool_use` is left with no result, or a `tool_result` survives whose `tool_use` was cut. teamclaude forwards the body as-is, Anthropic returns the 400, and because it's non-retryable every follow-up in the session fails too. The session is wedged until the client is rewound or abandoned.

I hit this repeatedly with GUI agents (Paseo, and the Claude Code CLI after an interrupted turn) routing through teamclaude. It's the same class of bug the OpenCode ecosystem works around client-side, but agents that point straight at the proxy have no such guard. Since teamclaude is the one place every client funnels through, fixing it here covers all of them at once.

## The change

`src/tool-pair-sanitize.js` adds `sanitizeToolPairs(body, url, contentType)`. It runs in `forwardRequest` right before the existing `account_uuid` and model-map rewrites, so it reuses the body buffering and the `content-length` realignment that are already there.

It only ever removes provably-unpaired blocks:

- a `tool_use` with no `tool_result` anywhere in the body, or
- a `tool_result` whose `tool_use` is gone.

It never fabricates tool output the model would reason over, which keeps it safe by construction. A few details worth calling out:

- A well-formed body is returned as the **same Buffer instance**, so it's a genuine no-op on the hot path (the `sendBody !== body` check already in place skips the content-length update).
- Only `/v1/messages` (and `/v1/messages/count_tokens`) JSON bodies are inspected. Anything else, non-JSON, or unparseable is passed through untouched.
- If pruning empties a whole message, same-role neighbors are coalesced so roles still alternate (Anthropic requires that), rather than leaving two `user` or two `assistant` turns adjacent.

## Tests

`test/tool-pair-sanitize.test.js` (node:test) covers the reported tail-orphan case, a dangling `tool_result`, an orphan alongside real content, the same-role coalescing path, the same-Buffer no-op, and the non-messages / non-JSON / count_tokens paths.

`npm run lint` is clean and `npm test` passes for everything I touched. One unrelated test (`test/server-listen-error.test.js`, "server exits cleanly when the port is already in use") fails on my machine, but it fails identically on a clean `master` checkout with this change stashed, so it's a pre-existing environment flake, not from this PR.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -5,6 +5,7 @@ import { mkdir } from 'node:fs/promises';
 import { join } from 'node:path';
 import { ensureCerts, createConnectHandler } from './mitm.js';
 import { patchAccountUuid } from './account-uuid-rewrite.js';
+import { sanitizeToolPairs } from './tool-pair-sanitize.js';
 import { parseRequestModel, parseAdvisorModel } from './account-manager.js';
 import { TopLevelFieldFinder } from './model.js';
 import { BodyWriter } from './request-log.js';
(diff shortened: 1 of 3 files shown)


viiqswim:
Pushed a follow-up fix. The first version paired tool blocks across the whole body: it kept a `tool_use` as long as a matching `tool_result` existed anywhere. That misses a real case. Anthropic enforces the rule positionally, the result has to be in the immediately following message, and a compaction can *separate* a pair (both blocks survive but other messages slip between them). Whole-body pairing saw the id still present and forwarded the orphan, so the 400 came back.

The sanitizer now checks the true neighbor: a `tool_use` is kept only if the next message answers it, and a `tool_result` only if the message right before it grounds it. Pruning runs to a fixed point because dropping a turn can expose the next orphan, and same-role neighbors are coalesced so roles still alternate. Well-formed bodies still return the same Buffer, so it stays a no-op on the hot path. Added a separated-pair regression test and a cascade test; `npm run lint` and the suite are green. I reproduced the original failure and the fix live against a running proxy.

Issue #84, opened 2026-07-06
Storming


roderik:
I was doing WAY WAY too much, think 30 agents at full tilt. One account ran out, so it switched, and immediately threw them all against the next account, which got throttled (also with full context so 30% gone by that time). So it jumped again, throttle again, etc etc. 

When switching it should do some ramp up to prevent this when it suddenly retargets.

jfarre20:
or a way to take multiple requests and span across accounts simultaneously, trying to match prev requests to same account for kv caches?

MagicalTux:
Generally when I have a lot of claudes working, their requests take 2~5 minutes each, which means when an account switchover happens they don't all request at the same time. I'll see to add a storm detection / ramp up.

PR #114, opened 2026-07-19
fix(routing): stop reporting a healthy pool as fully exhausted

songwunil:
## Summary

TeamClaude could return `All N accounts exhausted` (HTTP 429) even when a healthy
account was available, and it forwarded a live upstream `401` straight to the
client. This PR keeps authentication failure, quota exhaustion, transient
throttling, route exclusion, and per-request failure as **separate** states, so
the proxy selects a requestable account whenever one exists and only reports the
condition that is actually true.

Reproduced on current `master` (not just an installed copy) before fixing.

## The bug, concretely

Given a pool of `[healthy, auth-rejected, quota-exhausted]`:

- **Before:** the request often ended in the terminal "no account" branch, which
  had a single message and always returned `429 rate_limit_error` with
  `All 3 accounts exhausted` -- even though one account was perfectly usable, and
  even when the real blocker was an expired credential or a transient per-minute
  429, not usage limits.
- A live `401` from upstream was passed back to the caller instead of isolating
  the one rejected account and retrying on a healthy one.
- After `enable` / `priority` / `reload`, config and credentials were synced but
  fresh usage was never consumed, so a stale `exhausted` flag could outlive a
  repaired account until the next natural probe.

## Root causes (all confirmed on current master)

1. The terminal no-account branch in `server.js` collapsed authentication, quota,
   temporary-throttle, and request-local exclusion into one synthetic
   `All N accounts exhausted` 429.
2. A live upstream `401` was neither quarantined to the rejected account nor
   failed over to a healthy one.
3. `enable` / `priority` / `reload` did not revalidate against the zero-spend
   usage endpoint, so a stale hard `exhausted` state was not retired even after
   the account became healthy again.

## What changed

**`src/account-manager.js`**
- `getAvailabilitySummary(exclude, model, advisorModel)` -- new; returns
  mutually-exclusive counts:
  `available / authentication / quota / temporary / disabled / route / excluded`.
- `markAuthenticationFailed(index)` -- account-local auth quarantine (never
  poisons the rest of the pool, never collapses into quota).
- `applyUsageData(...)` retires a stale `exhausted` state once a live usage
  response is under the quota threshold (the quota numbers still gate selection
  normally).
- Re-enabling an account clears a cached `error` / `throttled` / `exhausted`
  state (previously only `error`).
- `ensureTokenFresh(...)` now returns a structured `{ ok, classification }`
  (`not-applicable | fresh | refreshed | authentication | transient`).

**`src/server.js`**
- The terminal branch now emits an accurate response:
  `429 all_quota_exhausted`, `429 temporary_rate_limited`, or
  `503 mixed_pool_unavailable`, with a message that names the real counts.
- A live `401` gets one coalesced OAuth refresh + single retry; if it still fails
  (or it is an API-key 401), only that account is quarantined and the request
  fails over within the **existing** bounded retry walk.
- `POST /teamclaude/reload` accepts an optional `x-teamclaude-revalidate-index`
  header to revalidate exactly one account against live usage.
- The transient rate-limit `429` behavior (#84: pause + retry the same account,
  never rotate the burst) is deliberately **unchanged**.

**`src/prober.js`**
- `probeAccount(...)` returns `{ ok, classification }` and quarantines a repeated
  provider `401` without logging any credential.

**`src/index.js`**
- `enable` / `priority` / `reload` revalidate the target against live provider
  usage before re-selecting, and surface a non-requestable target (non-zero exit).

## Tests

Adds `test/quota-routing-availability.test.js` (23 tests, `node --test`):

- one healthy account is selectable beside two exhausted accounts;
- healthy selection survives a mix of one auth failure + one exhausted account;
- lower numeric priority preempts, and an auth-error priority target falls back;
- a stale `exhausted` state is cleared by live under-threshold usage, then the
  target is re-selected after `enable`;
- a real usage rejection rotates and preserves quota classification;
- an all-quota pool returns `all_quota_exhausted`; an all-temporary pool returns
  `temporary_rate_limited` (never mislabeled as quota);
- a mixed auth+quota pool returns `mixed_pool_unavailable`, never a fleet-wide
  quota claim;
- concurrent admission serializes a live lease without poisoning availability;
- a live `401` is isolated and the request fails over to a healthy account, with
  the healthy pool count preserved;
- concurrent `401` recovery coalesces to a single OAuth refresh and releases
  every lease;
- credential sentinels never appear in logs.

`npm test` (full suite) and `npm run lint` both pass locally.

## Compatibility

- `POST /teamclaude/reload` still accepts and returns the old numeric `added`
  form; the new response fields are additive.
- New `AccountManager` methods are additive; no existing signature changed in a
  breaking way (`ensureTokenFresh` previously returned `undefined`; callers that
  ignore the return value are unaffected).
- No config-schema or CLI-flag changes.
- One behavior change worth calling out: `enable` / `priority` now set a non-zero
  exit code when the running server reports the target is still not requestable
  (previously it exited 0 silently). This surfaces a real failure; a script that
  assumed exit 0 on every enable/priority should be aware.

## Risk / non-goals

- No second selector and no new router -- the fix stays inside the existing
  selection and retry paths.
- Bounded everywhere: one forced OAuth refresh per account per request; failover
  bounded by the existing account-count retry walk. No infinite retry, no
  fail-open (a fully-unavailable pool returns a structured 5xx/429, never a
  known-bad account), no account starvation, no priority lock, no 401
  propagation.

### DIFF
--- a/src/account-manager.js
+++ b/src/account-manager.js
@@ -473,6 +473,49 @@ export class AccountManager {
   }
 
   /**
+   * Explain why each account is or is not requestable without exposing account
+   * identities. The server uses these mutually-exclusive counts to avoid
+   * collapsing authentication failures, quota exhaustion, transient throttles,
+   * routing exclusions, and per-request failures into one synthetic message.
+   */
+  getAvailabilitySummary(exclude = null, model = null, advisorModel = null) {
+    const counts = {
+      available: 0,
+      authentication: 0,
+      quota: 0,
+      temporary: 0,
+      disabled: 0,
+      route: 0,
+      excluded: 0,
+    };
+
+    for (const account of this.accounts) {
+      if (account.disabled) {
+        counts.disabled++;
+        continue;
+      }
+      if ((model && !this._routeAllows(account, model))
+          || (advisorModel && !this._routeAllows(account, advisorModel))) {
+        counts.route++;
+        continue;
+      }
+      if (!this._isAvailable(account, model, advisorModel)) {
+        if (account.status === 'error') counts.authentication++;
+        else if (account.status === 'exhausted' || this._isNearQuota(account, model)) counts.quota++;
+        else if (account.status === 'throttled') counts.temporary++;
+        else counts.quota++;
+        continue;
+      }
+      // Preserve persistent state evidence above. `excluded` is only for an
+      // otherwise-requestable account that failed earlier in this request.
+      if (exclude?.has(account.index)) counts.excluded++;
+      else counts.available++;
+    }
+
+    return { total: this.accounts.length, counts };
+  }
+
+  /**
    * Normalize and store the configurable routing table. A route pins a set of
    * model globs to an exclusive set of accounts (and may override the governing
    * quota bucket). Called from the constructor and on config reload.
@@ -977,10 +1020,13 @@ export class AccountManager {
     const account = this.accounts[accountIndex];
     if (!account) return;
     account.disabled = disabled;
-    if (!disabled && account.status === 'error') {
+    if (!disabled && ['error', 'throttled', 'exhausted'].includes(account.status)) {
       account.status = 'active';
       account.rateLimitedUntil = null;
-      console.log(`[TeamClaude] Account "${account.name}" re-enabled — clearing error state`);
+      account.throttledAt = null;
+      account.pausedUntil = null;
+      account.requalify = true;
+      console.log(`[TeamClaude] Account "${account.name}" re-enabled — clearing cached unavailable state`);
     }
   }
 
@@ -1009,6 +1055,14 @@ export class AccountManager {
     if (usage.sevenDayFable) {
       if (usage.sevenDayFable.utilization != null) q.unified7dFable = usage.sevenDayFable.utilization;
       if (usage.sevenDayFable.resetAt != null) q.unified7dFableReset = usage.sevenDayFable.resetAt;
+    }
+
+    // A successful usage response is live evidence. Retire the legacy hard
+    // `exhausted` state once the governing quota is below the live threshold;
+    // the quota values remain authoritative and still gate selection normally.
+    if (account.status === 'exhausted' && !this._isNearQuota(account)) {
+      account.status = 'active';
+      account.requalify = true;
     }
 
     // If we just learned this account's weekly window while probing, re-evaluate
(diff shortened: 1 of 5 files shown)


songwunil:
Closing this PR: it targets the wrong base for the production issue it was meant to fix, and the intended target already contains the fix.

**Wrong base.** This PR was opened against `KarpelesLab/teamclaude:master`, but the deployment that needs the fix tracks the fork `jung-wan-kim/teamclaude` (its own `master`, currently `35f2f9b`, Release 1.2.3). That fork is substantially diverged from KarpelesLab master (ahead 85, behind 77 at time of writing), so this diff cannot be cleanly or meaningfully applied there.

**The fork already fixes this failure class natively.** The failure this PR addresses — a healthy account exists, another is 401/invalid, another carries stale exhausted quota, yet the pool is reported fully exhausted — does not reproduce on the fork's current `master`. Verified by reading the fork's routing end to end and by a focused 11-test regression matrix that passes against the fork's pristine tree with **zero source changes**:

- a healthy account is selected while one peer is `error`/401 and another is over-threshold (unit + live proxy) — the terminal "all accounts exhausted" branch is unreachable while any account is requestable;
- a stale over-threshold state is retired once fresh usage proves capacity, and an expired window is swept lazily without a probe;
- `isExhausted` separates real quota exhaustion (`unified-status: rejected`) from a transient under-threshold 429, and an all-authentication-failed pool returns 401 rather than a rate-limit 429;
- enable/disable/priority changes leave no stale routing eligibility (a disabled account is never revived; a priority change routes immediately);
- concurrent OAuth refreshes coalesce onto a single in-flight refresh.

The fork implements all of this through its own use-or-lose selection, the `_isAvailable` gate, `isExhausted`-based 429 classification, `computeRetryAfter` (which already folds throttle expiry and every binding quota reset), and `_refreshPromise` single-flight — not through this PR's `getAvailabilitySummary` architecture. Porting that architecture onto the fork would replace working fork-specific code, so no replacement PR is warranted (per "do not manufacture an empty or regressive PR").

**Nothing is destroyed.** The regression matrix is preserved as a branch on the fork (`test/hue-quota-routing-stale-state-proof`), and this PR's branch and diff remain in history. If KarpelesLab's own base still exhibits the collapse this PR fixes, it can be reopened or retargeted for that base on its own merits — this closure is only about it not being the right vehicle for the fork's production issue.

PR #124, opened 2026-07-22
feat(proxy): forward plain-HTTP requests to third-party hosts

MagicalTux:
## Problem

Tools running alongside teamclaude that honor `HTTP_PROXY` (gh, curl, npm, …) send a **plain-HTTP** request to the proxy in **absolute form** — `GET http://host/path`. teamclaude treated that as an origin request and routed it through account selection to Anthropic, so the call came back as a bogus account error instead of reaching its target:

```
$ curl http://example.org/          # HTTP_PROXY -> teamclaude
{"type":"error","error":{"type":"rate_limit_error","message":"All 3 accounts exhausted. Retry in 60s."}}
```

HTTPS already worked — the CONNECT tunnel (`mitm.js`) blind-relays non-Anthropic hosts end-to-end. Only the plain-HTTP forward-proxy path was missing.

## Fix

Detect an absolute-form request URL in the base server and **blind-relay it to its target host** with the client's own headers — no account selection, no token injection, `content-encoding` passed through untouched (a transparent forward proxy). It's the plain-HTTP analogue of the CONNECT tunnel.

- Anthropic traffic is never absolute-form here (it's origin-form via `ANTHROPIC_BASE_URL`, or a CONNECT tunnel), so this only ever sees third-party hosts.
- Runs **after** the existing auth gate (loopback exempt, remote needs the proxy key), so it isn't an open relay.
- Hop-by-hop and `proxy-connection` headers stripped; `502` on an unreachable target instead of a hang.

## Tests

`test/http-forward-proxy.test.js` spins up a real `createProxyServer` + a target origin server and asserts: an absolute-form GET is forwarded to the target (with an accountManager whose `getActiveAccount` throws, proving Anthropic is never consulted), a POST body+method is forwarded, and an unreachable target yields `502` rather than hanging.

Full suite green (316), lint clean.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -74,6 +74,12 @@ export function createProxyServer(accountManager, config, hooks = {}, sx = null)
         return;
       }
 
+      // Forward-proxy request (HTTP_PROXY): an absolute-form URL means a tool
+      // running alongside teamclaude (gh, curl, npm…) is proxying plain HTTP to a
+      // third-party host. Blind-relay it there — Anthropic traffic is never
+      // absolute-form here, so this only ever sees third-party hosts.
+      if (/^https?:\/\//i.test(req.url || '')) { relayHttpForward(req, res); return; }
+
       // Status endpoint
       if (req.method === 'GET' && req.url === '/teamclaude/status') {
         const status = accountManager.getStatus();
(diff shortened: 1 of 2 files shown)


MagicalTux:
Closing per maintainer call: in practice everything proxied is HTTPS (handled by the CONNECT tunnel), so plain-HTTP forward proxying isn't a real use case and isn't worth the added surface. The only residual is that an absolute-form plain-HTTP request returns a misleading "accounts exhausted" 429 — a rare, clear-fail footgun, not a hang. Easy to reopen if that ever comes up.

PR #154, opened 2026-08-03
feat(codex): ChatGPT Codex accounts as a backend protocol

DIodide:
Adds ChatGPT Codex as a backend protocol, so a ChatGPT subscription can serve requests alongside Claude accounts.

This is deliberately different from the existing third-party backend support. That path works because DeepSeek and GLM publish Anthropic-compatible endpoints, so `upstream` + `modelMap` is enough. Codex speaks the OpenAI Responses API — different request shape, different SSE event model, different auth — so it needs real translation in both directions.

I've kept everything outside the forwarding path protocol-agnostic. Selection, rotation, priority, routes, `disable`/`enable`, `TC_ACCT` pinning, quota thresholds, status rendering and state persistence are untouched; a codex account is a peer in the same fleet, not a parallel code path. The one place that needed protocol awareness outside the wire layer was `accounts`, which was fetching an Anthropic profile for every OAuth account.

## What's here

- **`login --codex`** — its own PKCE authorization against `auth.openai.com`, so TeamClaude's credential is independent of the Codex CLI's. `import --codex` also works but copies `~/.codex/auth.json`, which means both hold one refresh token; OpenAI rotates it on every refresh, so whichever refreshes second is left with a dead one. The import path says so and points at login.
- **Request translation** — Anthropic Messages → Responses API: system prompt to a developer message, content blocks to input items, tool schemas to function declarations, thinking budget to reasoning effort.
- **Response translation** — a stateful Codex SSE → Anthropic SSE translator. The two event models don't line up: Anthropic needs content blocks opened and closed around a monotonic index, while Codex interleaves reasoning, text and tool calls freely.
- **Prompt caching** — the cache is scoped by a key derived from session, sub-agent and model, set both as `prompt_cache_key` and on the `Session_id` header, mirroring CLIProxyAPI's `ClaudeCodePromptCache`. Without it the backend falls back to echoing `Session_id`, which quietly meant *zero* caching for any client that sends no session header. Measured on a 3k-token prompt: a stable key caches 2816 of 3016 input tokens from the second turn on, a per-request key caches nothing. `cache_write_tokens` is also mapped onto `cache_creation_input_tokens`, which was being dropped.
- **Quota** — read from `x-codex-*` response headers and normalised onto the existing `unified5h`/`unified7d` fields, so rotation is predictive rather than waiting for a 429.
- **Token refresh** — codex accounts are excluded from the prober (it reads an Anthropic endpoint that rejects a ChatGPT token) and from keep-warm (their plan has no 5h session to hold open), so they get a small scheduled refresher instead. It checks often and refreshes only inside a 30-minute window, which works out to roughly one refresh per token.
- **Model discovery** — `/v1/models` answers with the Codex catalog, so Claude Code lists GPT models in its own picker when `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY` is set. Claude Code drops any listed id not starting with `claude-`, so ids are encoded into claude-prefixed ones and decoded before anything reads the model — the same encoding CLIProxyAPI uses (`EnsureClaudeModelIDPrefix`), so a setup written for one behaves the same on the other. The listing carries only what TeamClaude adds, since echoing Anthropic's catalog back duplicates every Claude entry in the picker; `modelDiscovery.includeAnthropic` restores the full list. `blockedModels` now filters the listing too — advertising a model the proxy would reject with a 400 puts a choice in the picker that fails the moment it is used. Selecting one just works: selection is protocol-aware, so a Codex model routes to a codex account and a Claude model never does. That also fixes the reverse case — a codex account with no `modelMap` used to be picked for `claude-*` requests and 400 on every one. `modelMap` is now only needed to make a codex account answer to Claude model names as a quota fallback. Only intercepted when a codex account exists, so an Anthropic-only fleet is untouched.
- **Observability** — the serving account and resolved model are recorded per session and exposed at `GET /teamclaude/status` under `sessions.byId`, and the TUI activity line shows `(claude-opus-5 → gpt-5.6-sol)`. This one is useful beyond codex: a DeepSeek or GLM account with a `modelMap` has the same blind spot today, and it's easy to split out if you'd rather take it separately.

## How it was tested

The translators were ported from CLIProxyAPI's reference implementation and **differential-tested against it**: a small Go harness calls its translators directly, and the JS port is diffed against them. 38 request cases and 30 stream cases match byte-for-byte. Those outputs are committed as golden fixtures, so the suite asserts against ground truth without needing Go or a CLIProxyAPI checkout.

Both translators matched on the first run, which I didn't trust, so I mutation-tested the corpus — 12 plausible porting mistakes, of which the first pass killed only 9. The three survivors were real gaps (stale terminal arguments overwriting streamed deltas, missing deferral of items behind an active tool call, and the stop-reason path where Codex names a tool finish that produced no tool block). Cases were added to close them; it now kills all 12.

Fixtures only get you so far, though, and running a real Claude Code session against the live backend found four things they structurally could not:

- Codex sends its SSE responses with **no `content-type` header**, so deciding "is this a stream?" from that header meant every response was buffered and returned as raw Responses-API events. My e2e mock had been setting the header — more polite than the real server, which is exactly how a mock hides a bug.
- An error response is JSON, not SSE, so forcing the streaming path on non-200 fed the error body to the translator, which emitted nothing. The client saw `400 (no body)` and the message explaining the failure was destroyed.
- Claude Code calls Anthropic-only endpoints (`bootstrap`, `oauth/usage`, `mcp_servers`) alongside `/v1/messages`; refusing them broke those features whenever rotation landed on codex. They now fail over.
- A thinking budget of ≤512 became effort `minimal`, which the backend rejects outright. The fixtures couldn't catch it because CLIProxyAPI clamps in a later pipeline stage they don't cover — a blind spot in testing function-against-function rather than pipeline-against-pipeline.

Verified live end to end: streaming, non-streaming, tool calls, a tool_result round trip, `modelMap` selection, and quota going from unknown to a recorded weekly bucket after one request.

Probing the backend for the model and reasoning-effort matrix also settled a disagreement between the only two sources available. The API's enum message advertises `minimal`, but sending it draws a model-specific rejection (`'minimal' is not supported with the 'gpt-5.6-sol' model`), so the documented enum is wider than any model accepts. Conversely the Codex CLI's model cache lists `ultra` as a supported level for sol and terra, but the API rejects it — that cache describes the CLI's menu, not the wire contract. The accepted set is `none, low, medium, high, xhigh, max`, which is what the clamp uses; `ultra` now lands on `max` rather than falling to the default.

All seven models the live catalog exposes are verified end to end through the proxy — each returning its own id in `message_start`, with no `modelMap` configured:

| Model | Description |
|---|---|
| `gpt-5.6-sol` | Latest frontier agentic coding model |
| `gpt-5.6-terra` | Balanced agentic coding model for everyday work |
| `gpt-5.6-luna` | Fast and affordable agentic coding model |
| `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini` | Previous generations |
| `codex-auto-review` | Review-oriented alias |

`npm test` is 581/582 and `npm run lint` is clean. The one failure is `test/server-listen-error.test.js`, which fails identically on an untouched `upstream/master` worktree here — environmental, not from this branch.

## Known limits

- Reasoning summaries aren't requested (matching CLIProxyAPI), so responses carry no thinking blocks. The thinking-block and signature-replay paths are fixture-verified but never exercised live, because the backend emits no reasoning events without an explicit summary opt-in.
- Only `/v1/messages` is translated; everything else needs a Claude account in the fleet.
- Auth impersonates the Codex CLI's client id and user agent, which is what makes a ChatGPT subscription usable here at all. Worth a deliberate decision on your side, and I'd understand a "no" on that basis alone.

Happy to restructure, split the observability commit out, or drop pieces — and equally happy to hear this is out of scope for the project, in which case it can live as a fork.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,11 @@ Notes and limits:
 - Reasoning effort is mapped onto the levels Codex accepts (`none`, `low`,
   `medium`, `high`, `xhigh`, `max`). `minimal`, `ultra` and `auto` have no
   equivalent and are clamped to `low`, `max` and `medium` respectively.
+- Prompt caching is scoped by a key derived from the session, sub-agent and
+  model, so it survives across a conversation's turns. On a 3k-token prompt
+  that caches ~93% of the input from the second turn onward. Cache reads and
+  writes are reported as `cache_read_input_tokens` /
+  `cache_creation_input_tokens`.
 - Reasoning summaries are not requested, so responses carry no thinking blocks.
 - A ChatGPT plan without Codex access (e.g. free) is rejected by the backend.
 
--- a/README.md
+++ b/README.md
@@ -144,6 +144,15 @@ Anthropic's catalog back would just duplicate every entry in the picker. Set
 `"modelDiscovery": {"includeAnthropic": true}` if a client needs the full
 list.
 
+Models in `blockedModels` are left out of the listing —
+advertising a model the proxy would reject with a `400` just puts a choice in
+the picker that fails the moment it is used. Patterns anchor at both ends, so
+trim older generations explicitly or with a wildcard:
+
+```json
+"blockedModels": ["gpt-5.4", "gpt-5.4-mini", "gpt-5.5"]
+```
+
 A `modelMap` is only needed for the other direction: making a codex account
 answer to Claude model names, so it can stand in as a fallback when Claude
 quota runs out. Pair it with a high `priority` so it stays a fallback:
@@ -437,7 +446,7 @@ When on, teamclaude routes each **new** session to the least-loaded eligible acc
 | `holdSeconds` | Maximum seconds to hold the connection when all accounts are exhausted, polling silently until one recovers (`0` = return 429 immediately, the default). `teamclaude run` raises `API_TIMEOUT_MS` automatically to match |
 | `distributeSessions` | Spread concurrent Claude Code sessions across equal-priority accounts, each session pinned to one account for cache reuse (`false` = quota-driven rotation only, the default). Session tracking/readout is always on regardless — see [Session-aware routing](#session-aware-routing-distributesessions-off-by-default) |
 | `eventLogging` | How to handle Claude Code's telemetry (`/api/event_logging/*`), which is high-volume activity-log noise: `hide` (default) forwards it but keeps it out of the activity log; `block` answers `200` locally without forwarding (no upstream round-trip); `show` forwards and displays it. Toggle live in the TUI settings screen (`g` → Event logging). |
-| `blockedModels` | Array of model glob patterns (e.g. `["*fable*"]`) whose requests are rejected with a fast, non-retryable `400` instead of being forwarded — avoids a model no account can serve getting rate-limited upstream and hanging the pipeline (issue #116). Edit live in the TUI settings screen (`g` → Blocked models). Empty (the default) blocks nothing. |
+| `blockedModels` | Array of model glob patterns (e.g. `["*fable*"]`) whose requests are rejected with a fast, non-retryable `400` instead of being forwarded — avoids a model no account can serve getting rate-limited upstream and hanging the pipeline (issue #116). Blocked models are also left out of the [`/v1/models`](#chatgpt-codex) discovery listing, so the picker never offers a model the proxy would refuse. Edit live in the TUI settings screen (`g` → Blocked models). Empty (the default) blocks nothing. |
 | `stormRamp` | Optional storm-control tuning (on by default) — see [Storm control](#storm-control-switchover-ramp-up). Object: `{ enabled, startConc, stepConc, stepMs, windowMs }` |
 | `sx.apiKey` | [sx.org](https://sx.org) API key. When set, TeamClaude auto-provisions a residential proxy (egress-IP 429 workaround). Absent/empty = off |
 | `sx.mode` | `always` (route all upstream traffic), `429` (direct, fail over to the proxy after a 429), or `off` (keep the key but don't use it). Defaults to `always` when a key is set |
--- a/README.md
+++ b/README.md
@@ -121,13 +121,28 @@ independent of the Codex CLI's. Prefer it over importing:
 teamclaude import --codex
 ```
 
-Codex models are offered directly in Claude Code's own model picker. Set
-`CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1` and Claude Code reads
-`/v1/models` from the proxy at startup; TeamClaude answers with the Claude
-catalog **and** every codex account's, so selecting `gpt-5.6-sol` is a
-first-class choice. Requests naming a Codex model route to a codex account
-automatically — no `modelMap` needed, and a Claude account is never picked for
-one (or vice versa).
+Codex models are offered directly in Claude Code's own model picker:
+
+```bash
+export ANTHROPIC_BASE_URL=http://127.0.0.1:3456
+export ANTHROPIC_AUTH_TOKEN=$(jq -r '.proxy.apiKey' ~/.config/teamclaude.json)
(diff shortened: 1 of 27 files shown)


MagicalTux:
Thanks for this — the engineering quality here is not in question. Differential-testing the translators against CLIProxyAPI, then mutation-testing the corpus when both passed first try and finding three real gaps, is a more rigorous validation story than most of this repo has. The four issues you found only by running a live session (the missing `content-type` on Codex SSE especially) are exactly the kind of thing fixtures structurally cannot catch, and calling that out is appreciated.

I am not merging it yet, and I want to be straight about why rather than leave it sitting silently.

This overlaps #95, which is my own in-progress multi-provider work, and the two take genuinely different architectures. #95 routes by host through MITM with one `AccountManager` per provider behind a router; this PR translates at `/v1/messages` and keeps a single pool. Both are defensible. They are not composable, so this is a design decision about where the provider seam belongs in this codebase, and that call is mine to make rather than something to settle by merge order.

The client-id point you raised in "Known limits" is the other half of it, and I am glad you surfaced it instead of burying it. It deserves a deliberate answer, not an implicit one.

So: leaving this open pending my architectural review against #95. Nothing is being asked of you right now — please do not rebase on my account. If I go with your approach, I would rather work from this branch than reimplement it. Two things I would likely take regardless of which way that lands:

- the observability commit (serving account + resolved model per session, exposed on `/teamclaude/status` and in the TUI) — you are right that `modelMap` has this blind spot for DeepSeek and GLM today, and it stands alone
- `blockedModels` filtering the `/v1/models` listing, for the same reason: advertising a model the proxy will 400 on is a bug independent of Codex

If you want to split either out into its own PR, I will review it on its own merits promptly. Otherwise I will come back to this thread once I have worked through the seam question.

thomasahle:
This looks great and would be very useful

MagicalTux:
Coming back to this with a decision, five weeks after I said I would. That wait is on me, and you deserved an answer sooner — particularly given how much of the work here was done to a standard the repo does not usually get.

I am closing it, and the reason is the architecture, not the quality.

I said in August that the seam question was mine to settle. Having now worked through it: the seam belongs at the **credential and destination**, not at the wire format. A second provider should forward its own protocol untouched, with TeamClaude choosing which account and which upstream — and nothing more. Translating Anthropic Messages to the Responses API and back means this project owns a bidirectional protocol bridge: a stateful SSE translator, a reasoning-effort clamp, a model-id encoding, and a standing obligation to track both vendors as they move. Every one of those is a place this repo would break when someone else ships a change. That is the part I do not want to carry, and it is the part your PR is mostly made of — through no fault of the implementation.

The other half is the client-id point you raised yourself under "Known limits". You were right to surface it instead of burying it, and my answer is that I do not want this project's default path to depend on presenting itself as the Codex CLI.

What that leaves in your favour, and I want it on the record rather than lost in a close:

- differential-testing the translators against CLIProxyAPI, then **mutation-testing the corpus** when both passed on the first run because you did not trust the result — and finding three real gaps that way — is a better validation story than anything else in this repository
- the four bugs you could only have found by running a live session, especially Codex sending SSE with **no `content-type`**, and your observation that your own mock had been "more polite than the real server, which is exactly how a mock hides a bug"
- clamping the thinking budget after discovering the documented enum is wider than any model accepts

Two pieces I flagged in August still stand on their own, and I would still take either as its own PR:

- the **observability commit** — serving account and resolved model per session, on `/teamclaude/status` and in the TUI activity line. A DeepSeek or GLM account with a `modelMap` has that same blind spot today.
- **`blockedModels` filtering `/v1/models`** — advertising a model the proxy will reject with a 400 is a bug regardless of Codex.

Neither depends on the translation layer. If you would rather not, that is completely fair after this wait, and no hard feelings either way.

For transparency about what happens next: #246 takes the passthrough approach described above, and I am reviewing it now. That is a decision about which seam this project can maintain, not a judgement that it is better work than yours.

PR #171, opened 2026-08-17
fix(status): show blocked models instead of reporting them available

nova28:
## The bug

With a blocklist configured, `teamclaude status` reports the blocked model as available:

```
"blockedModels": ["*fable*"]
```
```
Routing
  *fable*          → acct-a (auto)

  Fable    [██░░░░░░░░░░░░░░░░] 11% reset 6d3h
  Models   Opus ✓   Fable ✓          <-- but every Fable request 400s
```

Meanwhile the proxy is rejecting it correctly:

```
HTTP 400 in 0.00093s
{"type":"error","error":{"type":"invalid_request_error",
 "message":"Model \"claude-fable-5\" is blocked by teamclaude (matched \"*fable*\")."}}
```

## Why

Two independent gates decide whether a model can be served, and status only knew about one.

`modelRoutingLine()` builds the `Models` row from the account's weekly quota buckets alone (`unified7d`, `unified7dSonnet`, `unified7dFable`). That answers *"does this account have headroom left"*, which is a reasonable question and the wrong one here: the blocklist is checked in the request handler **before** an account is selected, so a request for a blocked model never reaches quota evaluation. A family with 89% of its weekly bucket free and a matching blocklist entry is unservable, and status said `✓`.

The blocklist wasn't just mis-rendered, it was **absent from the payload**. `getStatus()` never carried `blockedModels`, so neither the renderer nor `status --json` could see it:

```
$ teamclaude status --json | jq 'has("blockedModels")'
false
```

The TUI shows a blocked count, but a headless server has no way to surface the setting at all. Confirming a block meant reading the config or probing the proxy.

## The fix

- Report `blockedModels` in the status payload, read live from the shared config so the TUI's editor is reflected immediately (matching how the per-request gate already reads it).
- Add a `Blocked` header row listing active patterns, rendered only when non-empty. Its absence was what let a blocked model read as available.
- Mark a blocked family `⊘ blocked` in the `Models` row. The blocklist outranks quota, so it's checked first.
- Render a route whose globs are all blocked as `→ blocked` rather than listing accounts it can never reach.

### After, on the same server

```
TeamClaude status
Active       acct-a
Switch at    98%
Blocked      *fable*
...
Routing
  *fable*          → blocked (auto)

  Session  [██░░░░░░░░░░░░░░░░] 9% reset 4h41m
  Weekly   [█████████░░░░░░░░░] 50% reset 6d3h
  Fable    [██░░░░░░░░░░░░░░░░] 11% reset 6d3h
  Models   Opus ✓   Fable ⊘ blocked
```

Quota rows are deliberately untouched: `Fable 11%` still reports real headroom, which stays useful for knowing what unblocking would give back.

## Notes on the family match

`blockedModels` globs target concrete model ids while the status view reasons in families, so `claude-fable-5` never matches the literal string `Fable`. `findFamilyBlock()` accepts both spellings a block is naturally written in — the glob (`*fable*`) and a bare id (`claude-fable-5`) — and a lone `*` blocks everything. It's display-only and deliberately inclusive; the authoritative per-request gate in `server.js` is unchanged. A pattern naming no family (say `claude-3-*`) lights up no row, and the header still lists it verbatim.

`modelGlobOverlaps()` decides whether a route is fully shadowed. Exact glob intersection isn't decidable in general, so it compares literal cores in both directions.

## Tests

15 new assertions across `test/status-renderer.test.js` and `test/model.test.js`, covering: the `Blocked` row appearing and staying absent when empty, a blocked family not reading `✓` while it has free quota, blocking by concrete id as well as glob, unrelated families unaffected, and both route cases.

```
$ node --test test/status-renderer.test.js test/model.test.js
# pass 25
# fail 0
```

Full suite: 439/440. The one failure (`server exits cleanly when configured port is already in use`, `test/server-listen-error.test.js`) reproduces identically on unmodified `master` in my environment and is unrelated to this change.

`npx eslint` clean on all touched files.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -369,6 +369,10 @@ async function serverCommand() {
   // Expose reload to the proxy's control endpoint (works with or without TUI).
   hooks.reload = reloadAccounts;
   hooks.getStatusExtra = () => ({
+    // Read live from the shared config (not a startup snapshot) so the TUI's
+    // blocklist editor shows up in `status` immediately, the same way the
+    // per-request gate in server.js picks it up.
+    blockedModels: [...(config.blockedModels || [])],
     server: {
       startedAt: new Date(serverStartedAt).toISOString(),
       uptimeSeconds: Math.round((Date.now() - serverStartedAt) / 1000),
--- a/src/model.js
+++ b/src/model.js
@@ -52,6 +52,40 @@ function escapeRegExp(s) {
   return s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
 }
 
+// The `blockedModels` pattern that takes a model FAMILY out of service, or null.
+//
+// The blocklist is written against concrete model ids (`*fable*`,
+// `claude-fable-5`) but the status view reasons in families (`Fable`), so a
+// direct glob match is not enough: `claude-fable-5` never matches the literal
+// string `Fable`. Both spellings are checked so the two natural ways to block a
+// family light up the same row — the glob (via modelGlobMatches, which also
+// makes a bare `*` block everything) and a concrete id (via substring).
+//
+// Deliberately advisory: this drives display only. The authoritative gate is the
+// per-request check in server.js, which matches the real model id. A pattern
+// that names no family (say `claude-3-*`) simply lights up no row, and the
+// header list still shows it verbatim.
+// Do two model globs describe any model in common? Used to tell whether a route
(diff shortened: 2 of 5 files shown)


nova28:
Closing this in favour of a clean re-open from a fresh branch. Same change, no content differences. Will link the replacement here shortly.

nova28:
Replacement opened: #172

PR #178, opened 2026-08-19
feat(tui): burn-rate bar colouring and a column-aware name cell

lexfrei:
Three things, after #228 took the row-overflow half of this PR and #179 landed the width helpers.

The account name cell measures in display columns and cuts in UTF-16 units. A six-character CJK name survives `slice(0, 12)` intact, then `padEnd(12)` adds six more spaces, so the cell draws 18 columns where the row budgeted 12. A decomposed accent goes the other way and draws 11. This reproduces on master at `1d2ddb9` with none of this branch applied: at `W=100` a CJK row is 92 columns against an ASCII row at 86. Fixed with `rpad(truncate(name, nameW), nameW)`.

The name column then grows to the longest name when the terminal has room, floored at the current 12 so a narrow terminal renders exactly as it does today. It takes only the slack #228's budget leaves unspent, so the row invariants are untouched. The literal `40` in that budget is now `28 + NAME_MIN`, which is the same number and makes "the budget already reserves the name" checkable rather than a matter of trust.

Quota bars are coloured by burn rate: ahead of the pace the window can afford is red, on pace is green. At or above `switchThreshold` the bar is red whatever the pace, so a bucket the rotation already refuses reads red rather than green. Colouring by pace needs a window that is still running, so a bucket whose reset has passed takes the raw scale, the same fallback already used when there is no window at all.

19 new tests, 673 pass, up from 654.


MagicalTux:
Two things before merging: one design question, one mechanical.

**Design (for @MagicalTux):** with burn-rate colouring, a bucket at 98% with 99% of the window elapsed is green (`diff <= 0`), yet that account is unavailable for routing (`_isNearQuota` compares against `switchThreshold`); conversely a 5 h bucket at 20% ten minutes into its window is red. The status cell still says exhausted, but bar colour and eligibility now disagree. A hybrid keeps both signals: `ratio >= switchThreshold → red` regardless of pace, else burn-rate. `bar()` doesn't see the threshold today, so it needs one more argument.

**Mechanical:** #179 lands first. After it, `vw(a.name)` measures display width while `a.name.slice(0, nameW).padEnd(nameW)` is still UTF-16: a 6-char CJK name measures 12, keeps all 6 chars, pads 6 more → 18 columns and the row shifts. Use `rpad(truncate(a.name, nameW), nameW)` after rebasing.

The `columnLayout` arithmetic checks out against the row template (4 + routeSlot + 24 + 6 + fam×6).

thomasahle:
Heads-up so this isn't duplicated work: I opened #228 for the row-overflow half of this, after hitting the cut-off `F7` bar on a six-account Fable fleet at 86 columns.

I checked out this branch first and confirmed it does fix that — the countdown comes back. Two things led me to open a separate PR rather than wait or comment here:

1. The tag reserve stays unconditional, so the row still stops ~10 columns short of the edge on a fleet with nothing blocked (the reserve is spent whether or not any account draws a `⊘` tag). Same bug, opposite direction.
2. At ~70 columns with a blocked family there is no room for a third bar even at the minimum bar width, so the row overflows regardless of the budget — the bars have to be dropped there, which the budget alone can't express.

#228 is independent of the burn-rate coloring, so whichever lands first the other should rebase cleanly. I'd be glad to fold #228's layout part into this PR instead if you'd rather carry it here — your call, and no objection from me either way.

MagicalTux:
Heads-up: #228 (thomasahle) has merged and covers the overflow half of this PR — the row is now budgeted from every column it draws (route cells, per-bar labels, the ⊘ tag only when something is blocked, family bars dropped below the width where they fit at BAR_MIN), with a render-driven regression test. What remains distinct here is the name column growing to the longest name on wide terminals and the burn-rate colouring (design question above). Please rebase onto master; the width-aware name cell should use `rpad(truncate(a.name, nameW), nameW)` now that #179 is in.

lexfrei:
Both taken, rebased onto master, and the overflow work dropped since #228 covers it.

One correction on the design premise, because it changes what the clamp defends against. `'exhausted'` is read in `src/` and never assigned there. The only writers are tests, which set it on the account object directly. `'active'`, `'throttled'` and `'error'` are all assigned normally, so on a running proxy an account is never put into the status that sentence relies on. The divergence is real all the same, through `_isNearQuota` and the blocked-family tag, which is what the clamp reads.

The other direction you raised does not survive measurement, and I would rather show it than argue it. Same 5h bucket at 20%, no further spend, colour by elapsed time alone:

| elapsed | colour |
| --- | --- |
| 1 min | red |
| 10 min | red |
| 30 min | orange |
| 45 min | yellow |
| 60 min | green |

A colour that heals by waiting is reporting elapsed time, not risk. The rate estimate divides by elapsed, so its variance blows up near the window start, and spend arrives in lumps rather than a smooth flow. Ten minutes after a reset the bar has to be under 3.3% to read green, which is ordinary rather than an edge. If you want that closed, the shape is a minimum-elapsed guard below which burn rate does not engage and the raw scale runs instead. Say the word and it is one condition.

Worth naming, since the clamp is yours and the cost lands on you: red used to mean ahead of schedule and now means ahead of schedule or out of rotation, with nothing in the row separating them. Out of rotation is the more urgent fact so the merge is defensible, but the channel now carries two meanings.

On the stale-window point: the local renderer does call `refreshExpiredQuotas`, first thing in `_render`, so the bucket is already null there and the bar takes the no-data fill. Attach mode is a different story and the conclusion holds there. `RemoteAccountManager.refreshExpiredQuotas` is an empty no-op, `getStatus` copies the quota verbatim, and an idle headless server has nothing else that sweeps, so a 95% bucket past its reset renders green with a 95% label. That is pre-existing and this diff only changes which of two wrong colours shows, so I have left it alone rather than fold it in.

Two more, both pre-existing and outside this diff, in #234 together since they live in the same block.

The screenshot in the README predates 33 commits to `src/tui.js` and was already stale at the merge base. This branch adds to that drift rather than starting it.

PR #179, opened 2026-08-19
fix(tui): measure display width, not UTF-16 length, for table columns

lexfrei:
The table columns were sized with plain string length, which counts UTF-16 units, not terminal columns. A CJK char is one unit but two columns, a combining mark one unit but zero, so any account or org name with those broke the alignment and could cut a glyph in half. Now the width is measured with a small display-width table (East Asian wide and fullwidth ranges plus combining marks), walking the string by code point. ASCII is unchanged.

Not a live problem yet, since names are usually ASCII, but the old length count was wrong and would misalign the moment a non-ASCII name showed up. It does not resolve emoji ZWJ sequences, which stay counted per component, but names rarely carry those.


MagicalTux:
Pushed one commit on top: `truncate` now drops a wide glyph that would straddle the limit and returns w−1 columns (its comment says the caller pads), but `fitLine`'s too-long branch returned `truncate(s, w)` unpadded. The frame is repainted in place with no clear, so a w−1 line leaves the previous frame's last cell visible. `fitLine` now pads back to exactly w (exported, with two tests). Also noting for the follow-up in #178: the account name cell is still `a.name.slice(0, 12).padEnd(12)` (UTF-16), so a CJK account name still misaligns the row — this PR fixes `vw`/`truncate`/`rpad`, not that cell. Merging once CI is green.

PR #180, opened 2026-08-20
fix(proxy): respond on every error path instead of hanging the client

ak2k:
Three error paths in `server.js` catch a throw, log it, and return without writing anything to the socket. The client never gets a response and waits until its own timeout expires. One of the three is reachable from an ordinary request line, no hooks involved.

## 1. Malformed percent-escape in a `/tc-acct/` pin

The pin segment is percent-encoded by the client, and `decodeURIComponent` throws `URIError` on a malformed escape. The throw lands in the outer catch, which logs and returns:

```
$ curl -sS -m 6 -o /dev/null -w 'HTTP %{http_code} in %{time_total}s\n' \
    -X POST -H 'content-type: application/json' -d '{"model":"claude-opus-5","messages":[]}' \
    'http://127.0.0.1:8787/tc-acct/%/v1/messages'
curl: (28) Operation timed out after 6002 milliseconds with 0 bytes received
HTTP 000 in 6.002430s
```

`%zz` and a truncated `%E0%A4` behave the same way, while a well-formed unknown pin (`/tc-acct/ghost/...`) answers 404 in a few milliseconds.

The decode is now wrapped in a try/catch, and an undecodable pin falls through to the existing unknown-pin 404, since an undecodable pin and an unknown pin mean the same thing to the caller. The error message quotes the token as it arrived, because there is no decoded form to show. All the inputs above answer 404 in under 10 ms after the change.

## 2. Neither outer catch answers the socket

The 502 in `forwardRequest` covers the inner try only. The code above it (the egress hold, pin parsing, body buffering, the activity hooks) runs under a catch that logs and returns, and `createProxyServer`'s request handler has the same catch around the auth gate, the CSRF gate, the forward-proxy relay, and the status/reload/switch endpoints.

`getStatusExtra` is a hook the application installs, so this is reachable from a plain `GET`:

```
$ curl -sS -m 6 -o /dev/null -w 'HTTP %{http_code} in %{time_total}s\n' \
    http://127.0.0.1:8787/teamclaude/status        # with a status hook that throws
curl: (28) Operation timed out after 6002 milliseconds with 0 bytes received
HTTP 000 in 6.002208s
```

Both catches now send the same 502 that `forwardRequest` sends, when nothing has been written yet.

## 3. Throw after the headers are sent

The status endpoint serializes the hook's value after `writeHead`. If the hook returns something `JSON.stringify` rejects (a cycle, a `BigInt`), the throw happens with the 200 already sent. A 502 can no longer be written, nothing else ends the response, and the client hangs exactly as in the previous case. Measured on `master`, both inputs do.

For this case the catch destroys the response instead. `end()` would be wrong: it delivers truncated bytes as an apparently complete reply, and the client has no reason to retry. `forwardRequest` already handles its own errors with this writeHead-or-destroy pair; both outer catches now go through one shared `answerUnhandled` helper that does the same.

After the change, with the same three hooks:

```
  getStatusExtra throw   -> HTTP 502 in 0.003629s
  getStatusExtra cycle   -> curl: (52) Empty reply from server, 0.003077s
  getStatusExtra bigint  -> curl: (52) Empty reply from server, 0.003141s
```

The `headersSent` check matters in the other direction too: the inner `finally` calls `onRequestEnd` after the response has streamed, so a throw from that hook reaches the catch with the headers long sent, and an unguarded `writeHead` would raise `ERR_HTTP_HEADERS_SENT` from inside the catch.

## Tests

Five new tests: one in `test/account-pin.test.js`, four in a new `test/server-error-paths.test.js`. Each races its request against a 4 s timer, so a hang shows up as a failed assertion instead of a stuck run.

Four of the five fail on unmodified `master`, each by timing out:

```
✖ a /tc-acct/ pin with a malformed escape is answered, not left hanging (4002.770875ms)
✖ a throw above forwardRequest is answered rather than hanging the client (4007.947542ms)
✖ a throwing status hook is answered, not left hanging (4001.618458ms)
✖ a status hook returning an unserializable value closes the connection (4002.195959ms)
```

That last one asserts the connection was destroyed rather than merely that something came back, so an implementation that ended the response gracefully fails it too: swapping `destroy()` for `end()` in `answerUnhandled` turns it red with `the client read a complete 200 whose body was never written`.

The fifth test, `a throw after the response is streamed does not answer a second time`, passes on `master` and pins the new guard: it streams a real SSE response through `streamResponse`, throws from `onRequestEnd` afterwards, and fails if the recovery either answers twice or destroys the client's completed 200. Removing the `headersSent` check turns it red with `ERR_HTTP_HEADERS_SENT` as an unhandled rejection.

```
ℹ tests 530
ℹ pass 530
ℹ fail 0
```

525 tests before, 530 after. `npx eslint .` clean.

## Compatibility

Error paths only. Successful requests, resolvable pins, unknown pins, blocked models, 429s and upstream failures all take exactly the path they took before.

The new recovery does run after a response that completed: a throw from `onRequestEnd` reaches the catch with the reply already delivered. There it does nothing, because the response has ended and both arms decline, and the fifth test pins that. What a client can observe changing is a response that was still incomplete when the throw happened: it now gets a 502, or a closed connection if the headers were already out, instead of waiting for its own timeout.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -193,6 +193,9 @@ export function createProxyServer(accountManager, config, hooks = {}, sx = null)
       return forward(req, res);
     } catch (err) {
       console.error('[TeamClaude] Unhandled error:', err);
+      // The window above throws for real: `getStatusExtra` is a hook the
+      // application installs, and reload/switch reach the account manager.
+      answerUnhandled(res);
     }
   };
 
@@ -526,8 +529,33 @@ export function createProxyRequestListener({ accountManager, upstream, logDir = 
       }
     } catch (err) {
       console.error('[TeamClaude] Unhandled error:', err);
+      // The code above the inner try (the egress hold, the pin parsing, body
+      // buffering, the activity hooks) runs outside the 502 that guards
+      // forwardRequest, and the inner `finally` calls onRequestEnd after the
+      // response has streamed.
+      answerUnhandled(res);
     }
   };
+}
+
+/**
+ * The last response an outer catch can send. Three states:
+ *
+ *   - Nothing written yet: send a 502. Guarded on headersSent, because a second
+ *     writeHead raises ERR_HTTP_HEADERS_SENT from inside the catch.
+ *   - Headers sent, body unfinished: destroy. There is no status left to send,
+ *     and end() would present the truncated bytes as a complete reply.
+ *   - Response already ended: leave it alone, the client has its answer.
+ *
+ * `forwardRequest`'s own catch already carries the same pair of arms.
+ */
+function answerUnhandled(res) {
+  if (!res.headersSent && !res.destroyed) {
+    res.writeHead(502, { 'Content-Type': 'application/json' });
+    res.end(JSON.stringify({ type: 'error', error: { type: 'proxy_error', message: 'Internal proxy error' } }));
+  } else if (!res.writableEnded) {
+    res.destroy();
+  }
 }
 
 // Per-request https.Agent tunneled through sx.org — one-shot (no keep-alive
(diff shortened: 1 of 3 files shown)


MagicalTux:
Landed as part of #182 (which carries this commit verbatim). Thanks!

PR #184, opened 2026-08-20
fix(proxy): a DNS failure closes the connection instead of walking the fleet

ak2k:
teamclaude answers `rate_limit_error` when DNS is down.

A hostname has no per-account component, so an upstream name that will not resolve fails identically for every account. The failover path treated it as an account failure and tried the whole fleet:

```
before  4 of 4 accounts spent, 429 rate_limit_error "All 4 accounts exhausted"
after   1 of 4, connection closed
```

Both runs observed `getaddrinfo ENOTFOUND`, with the ambient outbound proxy disabled (behind one, the proxy resolves the name and the failure never arrives as a resolution error). Each walked account is a real upstream call charged to its bucket, and the log gets one line per account, so an outage also reads as several times its actual traffic.

## The classification

**Socket-scoped** codes describe this connection: `ECONNRESET`, `ECONNREFUSED`, `ETIMEDOUT`, `EPIPE`, and the five timeout codes (three undici, two of the proxy's own watchdogs). These stay unconditionally transient; the connection closes and the client retries. `EPIPE` is new to the set, the write-side sibling of `ECONNRESET`.

**Host-scoped** codes describe the host being dialled: `ENOTFOUND`, `EAI_AGAIN`, `EHOSTUNREACH`, `ENETUNREACH`, `ENETDOWN`. These are transient only while no untried account would dial a *different* host. An account may carry its own `upstream` for a third-party backend, and a name failure on the default host says nothing about that one, so that failover still works and is tested.

Codes are read from the error, its `cause` (global fetch puts the real error there), and its children (the happy-eyeballs dialer reports an all-addresses-failed connect as an `AggregateError` that can carry no top-level code).

## Why `ECONNREFUSED` is not in the host set

Nothing listening at `host:port` is arguably a property of the host, and an earlier version classified it that way. It was reverted. `ECONNREFUSED` was already unconditionally transient, so the conditional could only subtract: every gap in the condition becomes a live regression instead of an unfixed case. One gap is measurable today, since the condition scans the raw account list while selection applies its own availability check:

```
ECONNREFUSED unconditional   1 account spent, connection closed
ECONNREFUSED conditional     3 accounts spent, 429 rate_limit_error
```

(A disabled account carrying its own `upstream` is never selected, never enters the tried set, and satisfies "another host is available" for the whole request.) A test holds this at both levels: the unit assertion, and the fleet behaviour with a disabled account on its own backend.

## What this does not fix

The condition is a heuristic for whether failing over would reach somewhere new, and it is wrong in six known ways. None is introduced here.

<details>
<summary>Six limitations</summary>

1. **Accounts selection cannot choose still count.** Disabled, spent or excluded accounts never enter the tried set and satisfy the condition for the whole request. Measured above.

2. **A shared outbound proxy is invisible.** Accounts routed through one egress proxy reach the same real host whatever their `upstream` says.

3. **Hosts are compared as raw strings.** `http://localhost:1` and `http://127.0.0.1:1` read as two hosts.

4. **A pinned request counts accounts it cannot use**, so a host failure on it fails over into a path with nowhere to go and the client gets the 429 this change is meant to prevent.

5. **Behind an outbound proxy the code is gone before classification.** The proxy resolves the name and reports its own refusal, with no resolution code left to match.

6. **Disagreeing children are decided by evaluation order.** One `ECONNRESET` child makes an aggregate unconditionally transient where an `ENOTFOUND` sibling alone would have been conditional. Measured both ways.

</details>

The first four leave a host failure classified as worth failing over; the cost ranges from a fleet walk down to one wasted retry. The last two stop the condition being reached at all. Each costs less than the defect fixed here, and for `ECONNREFUSED` each would have been a new regression.

Separately: the flag chooses between closing and failing over, and does not reorder selection. On a mixed fleet the failover still walks same-host accounts in priority order before reaching one that dials somewhere new, so a mixed fleet saves less than the homogeneous `1 of 4` above.

## Tests

Eleven, in two files.

`test/upstream-error-classification.test.js` covers the predicate: socket codes stay transient with another host available, host codes do not, and codes are read from `cause` and from children. The aggregate case is asserted in the direction where reading the children changes the answer; the other direction is satisfied even by a version that never looks.

`test/unreachable-upstream.test.js` drives the real server and imports nothing this change adds, so it runs against the unfixed tree:

```
✖ an unresolvable upstream is not walked through the whole fleet
    one unresolvable host spent 4 of 4 accounts
✔ an unresolvable upstream still fails over to an account on a different host
✖ a fleet whose hosts are all unresolvable closes rather than reporting exhaustion
    two names that would not resolve were reported to the client as exhausted quota:
    429 {"type":"error","error":{"type":"rate_limit_error","message":"All 2 accounts exhausted. Retry in 60s."}}
✔ a refused connection is not walked through the fleet by an unselectable account
```

The two passing tests pin behaviour this change must preserve: failover across genuinely different hosts, and a refused connection staying unconditional. Each fleet test uses a fresh unresolvable name (no resolver cache), disables the ambient proxy through the existing test seam, and asserts it actually observed a resolution failure, so a resolver that hijacks unknown names fails loudly.

```
ℹ tests 536
ℹ pass 536
ℹ fail 0
```

525 before, 536 after. `npx eslint .` clean.

## Compatibility

Classification only. Nothing changes for successful requests, upstream 429s or 401s, or the codes already socket-scoped. `EPIPE` is the exception: a broken-pipe failure now closes the connection instead of falling through to failover.

One deliberate client-visible change: a host-scoped failure closes after one attempt instead of returning a 429 that blames quota for a network fault. A retrying client sees the same outcome sooner; a client honoring the 429's retry interval no longer waits on a quota state that was never real.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -722,6 +722,72 @@ function formatHeaders(headers) {
   return Object.entries(headers).map(([k, v]) => `  ${k}: ${v}`).join('\n');
 }
 
+// Failures that say nothing about the ACCOUNT, only about the socket. Retrying
+// can succeed where failing over cannot, and closing fast lets Node evict the
+// dead socket so the client's retry reconnects cleanly. EPIPE joins the set as
+// the write-side sibling of ECONNRESET.
+//
+// ECONNREFUSED sits here despite being arguably a property of the host. It is
+// already unconditionally transient, so making it conditional converts every gap
+// in that condition into a regression instead of leaving an unfixed case. One
+// such gap is measurable: a disabled account carrying its own `upstream` is
+// never selected, so it never enters `ctx.tried`, and the condition below then
+// reads as satisfied indefinitely. Wired that way, a four-account fleet spent
+// three accounts on a refused connection and answered rate_limit_error.
+const SOCKET_TRANSIENT = new Set([
+  'ECONNRESET', 'ECONNREFUSED', 'ETIMEDOUT', 'EPIPE',
+  'UND_ERR_CONNECT_TIMEOUT', 'UND_ERR_HEADERS_TIMEOUT', 'UND_ERR_BODY_TIMEOUT',
+  'TEAMCLAUDE_HEADERS_TIMEOUT', 'TEAMCLAUDE_BODY_TIMEOUT',
+]);
+
+// Failures that are a property of the HOST being dialled: name resolution and
+// routing. The hostname has no per-account component, so every account produces
+// the same failure, and walking the fleet spends an upstream call per account to
+// learn the same thing. The client is then told its quota is exhausted because a
+// name would not resolve.
+//
+// Conditional, because an account may name its own `upstream` for a third-party
+// backend. Where an untried account would dial a different host, this failure
+// says nothing about that one, and failing over is correct.
+const HOST_TRANSIENT = new Set(['ENOTFOUND', 'EAI_AGAIN', 'EHOSTUNREACH', 'ENETUNREACH', 'ENETDOWN']);
+
+/**
+ * Every error code a failure carries: its own, its `cause`'s, and its
+ * children's. Node's global fetch puts the real error on `cause`, and the
+ * happy-eyeballs dialer reports an all-addresses-failed connect as an
+ * AggregateError that may carry no top-level code at all, with the reason
+ * recorded once per address.
+ */
+function errorCodes(err) {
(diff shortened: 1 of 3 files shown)


ak2k:
Consolidated into #198, which carries this PR's commit with executor-eligibility scoping on top, to give the maintainer one read.

PR #187, opened 2026-08-24
feat(oauth): Add support for copy/paste authentication

notmgn:
Current `login` authentication flow spins up a local callback server, which only works if the browser and teamclaude are on the same box. On a remote server or in a container that's a dead end you end up logging in elsewhere and copying the credentials file over by hand.

This adds `login --token`: prints the auth URL, you log in from whatever browser you have, paste the code back. No listener, no localhost.

Along the way the token exchange and code parsing got pulled out into shared helpers, so the browser flow's existing paste fallback now handles the `code#state` format too.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,7 +8,7 @@ import { loadOrCreateConfig, loadConfig, saveConfig, atomicConfigUpdate, getConf
 import { installCrashHandlers } from './crash-log.js';
 import { AccountManager } from './account-manager.js';
 import { createProxyServer } from './server.js';
-import { importCredentials, loginOAuth, fetchProfile, refreshAccessToken, isTokenExpiringSoon } from './oauth.js';
+import { importCredentials, loginOAuth, loginOAuthWithPastedCode, fetchProfile, refreshAccessToken, isTokenExpiringSoon } from './oauth.js';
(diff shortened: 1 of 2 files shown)


MagicalTux:
Reviewed. The token path is right: the pasted code only ever goes to the existing token endpoint through `proxyFetch`, with the same client id, scopes and PKCE as the browser flow, and the refactor of `raceWithStdinCode` preserves its behaviour (URL → code/state check, raw → code) while adding `code#state`. Headless login is a real gap, so I'd like to take this. Two asks before merging:

1. `parseAuthCode` is pure and is the piece most likely to regress — export it and add unit tests for the four input shapes (callback URL, `code#state`, bare code, URL without `code=`) plus the state-mismatch throw. The PR currently has no tests.
2. Note in a comment that a bare code takes `expectedState` unchecked (PKCE still binds the exchange to this process's verifier, so it's acceptable, but it should be stated).

Optional: `--token` reads as "pass a token"; `--manual` or `--paste` would describe the flow better.

MagicalTux:
Merged as #239 (your commits, plus the parseAuthCode tests, rebased on master — the fork refused maintainer pushes so it went through a carrier branch). Thanks!

PR #188, opened 2026-08-24
feat(server): browser status dashboard at /teamclaude/dashboard

Frozen666:
## Motivation

The existing read surfaces — the TUI, `teamclaude status`, `teamclaude attach` — all require shell access to the machine running the proxy. For a proxy shared by several key holders, a URL anyone with a key can open is the friendlier way to check quota and rotation state.

## What this adds

`GET /teamclaude/dashboard` — one self-contained HTML page rendering what `teamclaude status` shows: per-account cards with quota bars (session 5h / weekly 7d, plus Sonnet/Fable buckets when present, with countdown and absolute reset time), status badges, current-account marker, usage totals, active-session counts. Auto-refreshes every 5s.

If the status payload carries a `clients` section (#186), a per-client usage table renders too — without it the section simply stays hidden, so this PR is independent of #186.

## Design notes

- **The asset is served before the key gate on purpose**: the page contains no data — everything it shows comes from `/teamclaude/status`, which stays gated and is fetched by the page's script with the key (asked once, kept in localStorage, re-prompted on 401). A browser address bar cannot send `x-api-key`, so gating the asset would 401 every remote browser without protecting anything. The route is exact-path GET only; a POST to the same path flows down the normal pipeline (covered by test).
- **Self-contained**: no external scripts, styles, or fonts — nothing to leak the localStorage-held key to, and it works on air-gapped deployments. All rendering uses `textContent`, so account/client names never reach innerHTML.
- Status timestamps arrive in both shapes (epoch-ms numbers for quota resets and account lastUsed, ISO strings elsewhere); the page parses both.
- Loopback stays key-exempt as everywhere else (a local browser sees status without entering a key).

## Tests

`test/dashboard.test.js`: the page is self-contained (no external src/href/@import) and fetches the gated status endpoint; GET serves HTML without a key while a POST to the same path is proxied normally. Full suite passes, lint clean.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -168,6 +168,16 @@ teamclaude help              # Show all commands
 
 ![teamclaude status output](assets/status-redacted.png)
 
+## Status dashboard (browser)
+
+`GET /teamclaude/dashboard` serves a self-contained HTML page rendering the same data as `teamclaude status`: per-account quota bars (session/weekly, plus Sonnet/Fable buckets where present), rotation state, active sessions — refreshed every few seconds.
+
+```
+http://localhost:3456/teamclaude/dashboard
+```
+
+The page is a static asset and loads without a key; the data does not — its script fetches `/teamclaude/status` with the proxy key, which it asks for once and keeps in the browser's localStorage (a 401 after a key rotation brings the prompt back). Loopback browsers are key-exempt as everywhere else. On deployments that put the proxy behind TLS this works remotely too: `https://your-proxy.example.com/teamclaude/dashboard`.
+
 ## Auto-update
 
 When TeamClaude is installed globally via npm, it self-updates in the background: it checks the npm registry at most once a day, and when a newer version is published it runs `npm install -g @karpeleslab/teamclaude@latest` and applies it on the next launch. The check runs after a `teamclaude run` session ends and when a headless server starts. A git checkout is never touched — update that with `git pull`. Run `teamclaude update` to update on demand.
--- a/src/dashboard.js
+++ b/src/dashboard.js
@@ -13,9 +13,48 @@
 // rendering uses textContent — status fields (account names, client names) are
 // operator/OAuth-derived, but they still never reach innerHTML.
 
+import { UNAVAILABLE_TEXT } from './status-renderer.js';
+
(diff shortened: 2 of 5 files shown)


MagicalTux:
Reviewed. XSS-wise this is clean (every DOM write goes through `textContent`, no external loads), and the page exposes nothing beyond the gated status JSON. Three notes:

1. The description and docs say a local browser sees status without entering a key, but `if (localStorage.getItem(KEY)) start(); else showKeybox();` plus `if (!v) return;` force a fresh browser to type something non-empty even on loopback. Poll first and show the keybox only on a 401.
2. The key sits in `localStorage` on the proxy origin, so add `Content-Security-Policy: default-src 'none'; script-src 'unsafe-inline'; style-src 'unsafe-inline'; connect-src 'self'; base-uri 'none'; form-action 'none'`, `X-Content-Type-Options: nosniff` and `Referrer-Policy: no-referrer` on the response.
3. This is the first unauthenticated 200 on the proxy port (the route sits before the key gate). It is a constant page, exact-match GET only, so the new exposure is a teamclaude fingerprint the 401 body already gives — but that is @MagicalTux's call. Note master now also has `teamclaude attach` (#170) for the dashboard-against-a-running-server case.

Frozen666:
Rebased onto master (1.1.14), and re-pointed at the fields 1.1.14 actually returns. One of those is a correctness fix rather than a cosmetic one:

- **Quota bars read `quota.scopedWeekly`.** `40e44d5` replaced the two dedicated Sonnet/Fable fields with a list learned from the usage payload, so hard-coding those two families drew an incomplete picture the moment upstream metered a third. The dedicated fields remain as a fallback for a payload that reports `seven_day_sonnet` without a `limits` array.
- **Account totals include `totalCacheReadTokens` / `totalCacheCreationTokens`.** `totalInputTokens` counts uncached input only, so the old total understated a Claude Code account by roughly the whole input side.
- **Surfaces `accounts[].unavailable`**, reusing `UNAVAILABLE_TEXT` from the status renderer rather than restating the map (it is now exported).

The two helpers this needs are pure, so they are exported and serialized into the page with `toString()` — one implementation, tested and served. `npm run lint` and `npm test` pass.

PR #189, opened 2026-08-24
fix(sessions): pin a session per weekly bucket, not once overall

ak2k:
A session holds one pin, so whichever model family routed last owns the whole session.

## The bug

Quota and eligibility are already decided per weekly bucket: an account whose Fable weekly is spent still serves Opus, and `_isAvailable` says so. Session affinity is the one place that reads a single answer out of a per-bucket world. A Fable request diverted off the pinned account re-pins the session there, and the next Opus request follows it onto an account nothing evaluated for Opus, whose Opus cache is cold.

Measured with a fixed schedule (one session interleaving Opus and Fable across four accounts, account `i`'s Fable weekly going spent after phase `i`, both trees seeing identical fleet state at every boundary): before, the Opus traffic re-homed twice, walking accounts it was never evaluated for; after, it holds its account for all nine of its Opus turns while the Fable half diverts independently.

## The fix

Key each pin by the bucket that governs the request, using `_weeklyBucketFor`, the same key selection already uses for eligibility. `recordSession` pins the bucket the request spent; `_selectForSession` looks up the executor's. `removeAccount` remaps session pins beside the route pins it already remaps, so both survive account removal symmetrically.

## Scope

`distributeSessions` gates the whole routing effect, and it defaults off. Flag off, 300+ interleaved requests with realistic durations across four families: account sequences byte-identical to master, per-session attribution 0 split / 0 unattributed on both trees.

What the flag does not gate is the readout: pins are recorded either way, and `perAccount` counts a split session on each account it is live on, once each, so it can sum to more than `active`. The invariant that holds: a routed, currently active session is counted on at least one account. One narrow case is disclosed in the body of the change: with two concurrent requests on different buckets, the account whose request ends last can read idle for up to the activity window.

## Tests

525 before, 541 after, 0 fail; `npx eslint .` clean. Reverting the pin keying with the new tests in place: 532 pass, 9 fail. Dropping only the end-of-request refresh: 538 pass, 3 fail. Dropping only the remap: 539 pass, 2 fail. Each red test names the behavior it pins.

## Notes

This is the base of a small stack: the expiry-routing PR and the accounting PR build on it. Merging out of order will conflict.


MagicalTux:
Real bug (a Fable diversion drags the Opus pin) and the `removeAccount` remap of session pins is a genuine fix on its own. One consequence the description does not state: the pin key is `_weeklyBucketFor(model)`, which returns `unified7dSonnet` for any Sonnet model whether or not the account reports that bucket. A session's first Sonnet request therefore has no pin and goes to `_pickLeastLoaded`, which counts the session's own fresh Opus pin as load — so with `distributeSessions` on and two equal accounts, an Opus+Sonnet session lands on two accounts by construction, not only on diversion, and counts as load on both. The README promises "pins it there". If the intent is one account per session unless forced apart, fall back to any available pin the session already holds before `_pickLeastLoaded`; that still passes your diversion test. Leaving the choice to @MagicalTux.

Issue #176, opened 2026-08-18
Rotation concentrates weekly burn on one account, so the pool starts refusing requests while siblings still hold weekly quota


johnnymo87:
> **Disclosure.** This issue was written by Claude (Opus 5) from telemetry
> collected off my own deployment. I read all of it, checked the claims against
> the source and the underlying data myself, and I am posting it because I agree
> with it. Any error in it is mine to answer for, not the model's.
>
> Please let me know if you agree with the diagnosis. If you don't, feel free to
> close this out — no hard feelings either way.

## Environment

- Observed on 1.1.11; selection path re-checked and **unchanged in 1.1.13**
  (`1342e92`). Line numbers below are v1.1.13.
- Measured on a 3-account OAuth Max pool, all at the default priority. (Our pool
  has since grown to five; every number in this issue is from the 3-account
  period, and the 5-account observations are kept separate below.)
- Evidence: 10 days of 5-minute quota sampling joined to per-request routing logs
  (~50k requests), plus a 7-day A/B against a patched build.
- **On how the refusals were counted.** The user-visible failure here is
  teamclaude's synthetic "all accounts exhausted" 429. Our own client sits behind
  a proxy that logs each of those and re-issues the request elsewhere, which is
  the only reason we can count them exactly and pair each one against the fleet's
  quota state at that instant. That re-issue is our local arrangement and not
  something this issue assumes anyone else has — read every count below as
  *"requests teamclaude refused"*, which is what it measures.

Related but not the same: #109 (closed via `distributeSessions`) was about
*concurrency* fairness across equal-priority accounts. This is about *weekly
quota* fairness. As far as I can tell that has not been filed.

## Observed behavior

Selection only leaves the current account when that account becomes
**unavailable**. `_select` (`:274`) returns `current` whenever `_isAvailable`
passes, and `_pickBestAvailable` (`:955`) ranks by priority → unknown-weekly-first
→ soonest weekly reset. Weekly *utilization* is never an input.

The result is that one account absorbs traffic until it hits a cap. Over our
baseline week the split of weekly quota consumed was **1.81 / 0.72 / 0.45
account-weeks** — 61% on a single account, where an even split would be 33%.

That would be merely untidy if the pool could still serve. It cannot, because the
5-hour bucket is per-account: an account that has been driven to its weekly cap
takes its 5h bucket out of the pool for the rest of the week, and the remaining
accounts have to carry the load at a lower aggregate ceiling.

[...]

Issue #175, opened 2026-08-18
A family (Fable/Sonnet) request is not gated by the shared unified7d bucket, so an account over its weekly cap keeps serving and overshoots further


johnnymo87:
> **Disclosure.** This issue was written by Claude (Opus 5) from telemetry
> collected off my own deployment. I read all of it, checked the claims against
> the source and the underlying data myself, and I am posting it because I agree
> with it. Any error in it is mine to answer for, not the model's.
>
> Please let me know if you agree with the diagnosis. If you don't, feel free to
> close this out — no hard feelings either way.

## Environment

- Observed on 1.1.11; code path re-checked and **unchanged in 1.1.13**
  (`1342e92`). Line numbers below are v1.1.13.
- 3-account OAuth pool (Max) at the time of measurement, mixed
  Opus/Sonnet/Fable traffic through a local proxy in front of teamclaude.
- Evidence: 10 days of 5-minute quota sampling (`/status`-derived) joined against
  per-request routing logs.

## Observed behavior

`_governingWeekly` (`src/account-manager.js:470`) returns the family-specific
weekly bucket **alone** whenever that bucket is reported:

```js
_governingWeekly(account, model) {
  const q = account.quota;
  const key = this._weeklyBucketFor(model);
  if (q[key] != null) return q[key];
  return key !== 'unified7d' ? q.unified7d : null;
}
```

So for a Fable request the availability decision consults `unified7dFable` and
never looks at `unified7d`. An account sitting at `unified7d = 1.00` with
`unified7dFable = 0.20` therefore still passes `_isNearQuota` (`:915`) and
`_isAvailable` (`:544`) for Fable, and keeps being selected.

That would be harmless if the buckets were independent. They are not: **spend on
a family model also increments the shared `unified7d` bucket.** We tested this
directly, regressing per-account `unified7d` deltas against Fable-only request
volume over a period with no non-Fable traffic on the account in question; the
coupling is positive and significant (bootstrap CI on the per-request increment
`[+1.14e-4, +5.21e-4]`, P ≈ 0.999). Family spend is metered twice — once in the
family bucket, once in the shared one.

The consequence is a one-way ratchet: once `unified7d` reaches its cap, family
requests are the only ones still admitted, and each one pushes `unified7d`
further past 1.00. We observed accounts reading above 1.0 on the shared weekly
bucket, which the code already anticipates (`:1079` notes utilization can
exceed 1) but never gates on for family traffic.

[...]

PR #190, opened 2026-08-24
feat(routing): expiry-pressure account selection

ak2k:
Stacked on #189 (per-bucket pins); merging out of order will conflict. Opt-in via config, default off, and with the flag off routing is byte-identical to the base.

## The problem

Selection never considers when quota expires while accounts are healthy. Reset-soonest ranking fires only at startup (#41 accepted that argument for the startup case) and at 98% exhaustion. At moderate utilization those moments never arrive, so a weekly window can expire mostly unspent on one account while traffic drains a longer-dated one. We measured a live fleet forfeiting 14 weekly points exactly this way.

## The mechanism

- **Pressure** per account per request model: governing weekly headroom divided by seconds until that same bucket resets. Value and clock always come from one bucket; dividing one window's headroom by another's clock scores an account on quota it does not have.
- **Top pressure band**: accounts within a tolerance ratio of the best pressure are placement-eligible; `distributeSessions` spreads within the band unchanged. Band membership is admission, not ordering.
- **Rollover-event preemption only**: a pinned session moves when its account's governing window actually rolls over, at most once per rollover, settled only when the session is quiet. Drain-driven band exit never preempts; a session draining its own pinned account is the policy working, and threshold preemption measured cache-thrashy in simulation with no quota gain.
- An account reporting utilization but no reset ranks by a lower bound (its spendable fraction over the longest possible window), so it is never preferred on the strength of a number nobody reported. Accounts reporting nothing rank in the top band; being used is how their window becomes known.

## Config

```json
"expiryRouting": { "enabled": true, "tolerance": 1.5, "preempt": true }
```

The key's spelling is provisional: #176 proposes a `routingStrategy` enum for the adjacent problem, and if the maintainer prefers one surface for both, this mechanism fits under it unchanged. The mechanism is settled; the spelling is negotiable.

## Evidence

- Flag off: two differentials against the base, zero divergences. Broad: 2,500 seeded fleets, 120 steps each, 148,574 picks compared. Targeted at the most-rewritten path: 192,000 forced entries through the session-reset switch plus 15,446 account removals, comparing both the returned account and the current pointer. A deliberately injected flag-off change trips the same probe 285 times at 400 seeds, so the zero is a measurement.
- 541 tests before (the #189 base), 623 after, 0 fail; `npx eslint .` clean.
- 31 targeted mutations (band ordering, rollover baselines, preemption guards, the bound), each reverted individually: every one turns at least one named test red; none survives.

## Compatibility

Default off and inert. With the flag on, per-account `pressure` appears in the status payload (observed only, never consumed by any display shipped here). No existing field changes.

The #175 fix (#191) and this PR both add one import name and one adjacent export in `src/account-manager.js` and `src/model.js`; if #191 lands first, this rebases the same day.


MagicalTux:
Read in full. Mechanically sound: opt-in, the flag-off paths reduce to today's priority → reset ordering, passive, no unbounded state, sane knob defaults, applied on reload. The open question is whether the project wants expiry-pressure routing at all — #176 and #212 ask the same thing and are unanswered, and this is ~2,300 lines (two new modules, ~1,200 lines of tests) to own. That is @MagicalTux's call, so I'm not merging on review alone.

If it goes ahead: (1) `_noteStuckRollover` logs "no eligible account can take that traffic" when `_selectNext` legitimately re-picks the current account because it still ranks best — distinguish the two; (2) `getActiveAccountFresh` is not threaded with `decision` (dead code today, but it becomes the one entry point that can never settle a rollover); (3) #195 and #211 have merged and touch the same regions, so a rebase is needed either way.

MagicalTux:
Reviewed, and I want this — but I am asking you to do the rebase rather than doing it myself, for a reason specific to how this PR is built.

## The review

The mechanism reads as settled. Deriving pressure with the value and the clock from the *same* bucket is the right call and the comment saying why (dividing one window's headroom by another's clock scores an account on quota it does not have) is the kind of thing that stops a later reader from "simplifying" it. Ranking an account that reports utilization but no reset by a lower bound, and letting accounts reporting nothing sit in the top band because being used is how their window becomes known, are both the conservative direction. Rollover-event-only preemption, with drain-driven band exit explicitly *not* preempting, is the part I would have argued for if it were not already there.

The evidence is the strongest I have seen on this repo. Two flag-off differentials with zero divergence across 148,574 picks is good; **validating the probe by injecting a flag-off change and confirming it trips 285 times at 400 seeds** is what makes the zero mean something rather than being an untested harness reporting success. Same for the 31 reverted mutations. That is the standard this change needed, given what it touches.

## Why I am not rebasing it for you

Master moved under you today, and not in a small way. Nine PRs landed, three of which collide directly with this:

- **#246** (Codex provider seam) took `getActiveAccount`'s fifth parameter for `provider`, and threads it into `_pickActiveAccount`. This PR takes that same slot for `decision`. Both need to survive.
- **#191** (the #175 fix) added `gatingUtilization` to `src/model.js` and rewrote `_governingWeekly` — the same function your pressure derivation reads.
- **#247** (per-account caps) rewrote the pinned-request branch in `forwardRequest` that your `ctx.decision` reset sits in, and added `capExceeded` to the same path.

I resolved comparable collisions in #191, #247 and #194 today. I could mechanically do the same here, and that is exactly the problem: this PR's value rests on a claim — *flag off is byte-identical* — that is established by your differential harness, not by reading the diff. A hand-merge of a 23-commit stack through an API-signature collision on the hot path is precisely the change most likely to break that property in a way review would not catch and your harness would. Landing my reconstruction of your work, unverified by the instrument that made it trustworthy, would throw away the reason to trust it.

So: please rebase onto current master and re-run the flag-off differential on the result. If it still reports zero, I will merge on that basis. You said in the description this rebases the same day if #191 lands first — it has, along with the other two.

Two smaller notes for when you do:

- `expiryRouting` as a spelling is fine by me for now. #176 is still open and unresolved, so if a `routingStrategy` enum ever happens it can absorb this; I would rather not block a working mechanism on naming a surface that does not exist yet.
- The two fixture commits at the base of your stack (`3310b27`, `83760c8`) are already on master as #260 — they will drop out or go empty on the rebase, which is expected rather than something going wrong.

Thanks for the care in this one. It is a lot of work and it shows.

ak2k:
Saw your review after I pushed. Rebasing onto current master now with the three collisions you named (#246's `provider` slot, #191's `_governingWeekly`, #247's pinned branch) and re-running the flag-off differential on the result; the numbers go here when it is done.

The head pushed just before your review is still on v1.1.14, so #195 and #211 are already underneath it, and the first two commits are the same fixture re-dates you merged as #260; they drop out on the rebase.

On the other two points from 2 September:

1. The held-rollover log line now says which case it was. `none-eligible` means nothing else could take the traffic. `ranks-best` means something else could have, but the account already serving it still ranked best. That comes from the eligibility check rather than from the re-rank result, so a one-account fleet always reads `none-eligible`, and an account barred by an upstream rejection (#230) counts as ineligible too.

2. The decision threading through `getActiveAccountFresh` is gone, and so is the in-flight bookkeeping behind it. A rollover is measured against the reading the last request left behind. When a preemption moves a session, the destination is adopted right away. The roll that displaced it is kept alongside so a fail-back gets it back, and it is dropped once a later request finds the pin where it was left. Nothing waits for the fleet to go quiet.

MagicalTux:
Merged as #282 — rebased onto current master and landed. Closing this in favour of that, since the work is yours and the history is preserved in the squash commit's co-authorship.

Two things your branch caught that I want on the record, because both were real and neither was obvious:

**Your `a chain answers the roll that started it` was describing a live defect, not an obsolete expectation.** #271 gave a rate-limit 429 one bounded failover hop, and that hop excluded only the accounts a request had *tried* upstream. An account a rollover preempted off was never sent a request, so it was absent from that set — and the hop handed the request straight back to the window the rollover had just moved it off, inside the same request, with nothing in the log to say so. Selection now reports what it deliberately routed away from and both hops honour it. Your test is what found it.

**The `_governingWeekly` collapse would have reverted #175.** Folding it into `_governingWindow().utilization` reads as obviously right — one resolution, no drift — but the two answer different questions: the gate takes the higher of the family bucket and the shared weekly, while the ratio must take value and clock from one bucket, as your own `_governingWeeklyReset` comment insists. `_governingWindow` returns the family bucket alone for a family model, so the gate would have gone back to letting an account at `unified7d` 1.00 keep serving Fable. They are kept separate with the reasoning written in place.

Three fixture updates were needed where #246, #271 and #277 had moved under you: the `a rate-limit 429` arm of `SAME_ACCOUNT_RETRIES` (no longer a same-account retry), a vestigial fifth argument that #246 turned into a `provider` collision, and the chain fixture needing a fourth account now that the hop spends one more.

The measurement standard you set — validating the differential harness by injecting a change and confirming it trips — is the thing I would most like to see more of here. Thank you for it.

ak2k:
Thanks for carrying it over the line. The `_governingWeekly` resolution reads correct — keeping it out of `_governingWindow().utilization` holds #175 on one side and the one-bucket pricing rule on the other, with the reason written where they diverge. Spelling out `expiryRouting: { enabled, tolerance, preempt }` and its default in the #176 and #212 closes puts it where anyone hitting the symptom will look.

PR #191, opened 2026-08-24
fix(quota): gate a family model on the higher of its own bucket and the shared weekly

ak2k:
The diagnosis, measurement, and proposed rule here are johnnymo87's, from #175. This implements their offer; read #175 first. Refs #175. If they would rather land their own patch, say so and this closes without hard feelings.

## The bug

`_governingWeekly` returns the family bucket alone whenever it is reported, so a Fable request on an account at `unified7d = 1.00` with `unified7dFable = 0.20` keeps being selected. #175 measured that family spend meters into both buckets, so this is a one-way ratchet: once the shared weekly caps, family requests are the only ones still admitted and each pushes it further past 1.0. We then watched a production account do exactly this at 99% shared utilization.

## The fix

The governing weekly for a family request is the higher of the reported family bucket and the reported shared bucket. `null` still means unreported, never zero; behavior is unchanged when the governing key already is `unified7d`; `_modelWeeklyExhausted` deliberately stays family-only, since "can this account serve this family at all" is a different question from "is it near any cap". The status `Models` row and the TUI tag read the same gate, so the display agrees with routing once the gate is fixed; on stock they only agreed because both shared the defect.

The recovery time beside a blocked mark is the latest blocking weekly bucket's reset: a block produced by a maximum clears only when every blocking bucket has rolled.

## Behavior change

Deliberate, and worth a release note: an account near its shared weekly cap becomes unavailable for family models sooner. That is the bug fix, not a regression, per #175's own framing.

Measured against stock: 5,000 seeded scenarios (2 to 5 accounts, quota levels drawn across the threshold, fixed 40-request mixed schedules), walking both trees in lockstep. 405 diverge, 4,595 identical, zero cases where this branch served a request stock refused. Every divergence falls into two classes: the stricter gate barred the account stock chose (358), or every account is now barred and the one revalidation probe was already spent (47). On the weekly buckets the family gate is now at least as strict as the shared one: over all 81 combinations of the two bucket levels, zero cases where an account can serve Fable but not Opus.

## Tests

525 before, 544 after, 0 fail; `npm run lint` clean. #175's six-row matrix is covered one test per row, plus the null cross-product, the probe ranker inheriting the fix, `_modelWeeklyExhausted` not inheriting it, the reset staying with the governing window, and six display cases through the real renderer and TUI row. With the fix reverted and the new tests in place: 536 pass, 8 fail. Upstream's own suite does not cover this path: the same reversion with the new file parked passes 525/525.

<details>
<summary>Limitations</summary>

1. A probe can still be aimed at an account whose shared weekly is spent (deliberate: `_modelWeeklyExhausted` stays family-scoped).
2. The soonest-reset tiebreak still ranks by the family window.
3. The recovery time considers the two weekly buckets; an account blocked longer by the five-hour bucket shows the weekly clearing time. Pre-existing on stock, occurs only while the weekly window is within about five hours of rolling; closing it is gated differently and out of this change's scope.
</details>

This collides with #190 and the #192 stack above it in `src/account-manager.js` and `src/model.js` (one import name and one adjacent export each); whichever lands second rebases the same day.


MagicalTux:
Gate logic checked against master's bucket names (`unified7dFable` / `unified7dSonnet` / `unified7d`, route `bucket` overrides, null = unreported): it implements #175's rule exactly, and the status/TUI display paths follow the same function. The open question is policy, not code. #175's own evidence is that upstream keeps serving Fable while `unified7d ≥ 1.0` (`7d_oi` = "overage included"). If that is served within the plan, this gate refuses requests upstream would serve; if it is billed overage, this is a cost guard and the right default. @MagicalTux knows the plan semantics — flagging for that decision. Trivial conflict with #190 in model.js.

PR #192, opened 2026-08-24
feat(usage): record what a request cost, per session and per weekly family

ak2k:
Stacked on #189 and #190; review those first, and note the diff here spans all three commits, only the last of which is this change. Merging out of order will conflict. Rebased 2026-08-24 so the accounting sits on the session-pin structure it measures. It inherits #190's collision with #191 (one import name and one adjacent export, in `src/account-manager.js` and `src/model.js`); if #191 lands first, #190 and this rebase together, the same day.

The proxy reads one field out of every usage object and discards the rest. Measured over 873,012 usage objects from 1,540 local sessions: the counted field (`input_tokens`, median 2) carries 0.05% of the input side; the two discarded fields (`cache_read_input_tokens`, median 199,824, and `cache_creation_input_tokens`) carry 99.95%.

## What this adds

Capture `input_tokens`, `cache_read_input_tokens`, and `cache_creation_input_tokens` per session, per weekly family bucket, and publish them in the status payload (`sessions.tokens`, per-bucket and per-account splits). Instrumentation only: nothing here reads the numbers back into routing, and no selection path changes. It gives an operator visibility into what sessions actually cost, and any future load-aware feature a measured foundation instead of a guess.

## The one subtle part

Streaming responses report usage as a `message_start` plus cumulative `message_delta`s. The deltas are merged and recorded once, in a `finally`, never re-added per event: a stream reporting `input_tokens` 12 at start and a cumulative 10,682 in its delta records 10,682, not 10,694. A stream aborted mid-flight still records the input it already spent, because only the `finally` runs. Both properties have direct tests driving the real proxy over a socket.

## Routing unchanged, measured

Both `distributeSessions` arms, seeded mixed-model schedules (four families, 90 to 108 requests per family per run), recording live on this branch throughout: 400,000 picks compared, zero differing, 349,050 usage reports recorded while comparing. A negative control (a branch copy whose recorder also nudges quota by 0.001) trips the same probe 1,896 and 3,203 times on the two arms, so the zero is a measurement, not an artifact.

## Tests

525 before, 561 after, 0 fail measured against master; on the rebased branch, 623 before and 659 after. `npx eslint .` clean. No existing test file is touched. 18 targeted mutations (field discarded, count frozen, wrong session charged, delta re-added per event, merge guard removed, published split aliased, record moved out of the `finally`, and others), each reverted individually: every one turns at least one named test red. The full reversion with new tests parked passes upstream's suite 525/525, which measures that no existing test covers this path.

## Compatibility

Additive fields only; nothing existing changes shape. Counters are since-process-start (`restoreQuotaState` restores quota, not usage). The per-session and per-account splits are asymmetric by design (six figures per family on the session side, two on the account side); the payload documents which.

<details>
<summary>Limitations</summary>

1. Counters reset on process restart.
2. The MITM path records normally in rewrite mode; tunnel mode sees no usage objects to record.
3. A session id reused after full expiry continues the old record on stock's session lifecycle.
4. The account-side split is coarser than the session side, deliberately.
5. Corpus figures above are from local transcripts of one deployment; the 99.95% split is corpus-local, with the command in the commit message.
</details>


ak2k:
Rebased onto #190 so the accounting sits on the session-pin structure it measures; the earlier conflict with #189 is gone. Content otherwise unchanged apart from the integration itself.

MagicalTux:
Additive and clean: nothing new is persisted (`saveState` only writes `PERSISTED_QUOTA_FIELDS`), per-session token maps live on the session record and are swept with it, `recordTokens` never creates a session, and the SSE last-wins merge of `message_start`/`message_delta` is right. It depends on #190 only for the `_live()` helper, not functionally — if #190 is not taken, please rebase onto #189 or master. Note the functional overlap with #194 (header-driven per-session attribution); the maintainer should pick one model, and this one has the cache-token accounting.

MagicalTux:
Merged as #241: your commit rebased onto master on top of #189 (without #190 — the `_live()` helper was added on its own) and merged with #186's per-client usage path. Thanks!

PR #193, opened 2026-08-25
fix: quarantine OAuth-entitlement-denied accounts

aidanbunch-ctx:
## Why

TeamClaude can repeatedly select an account whose organization refuses OAuth. The observed upstream response is a structured `403` with:

```text
error.details.error_code = oauth_not_allowed_for_organization
```

In the measured incident, 57 of 333 Fable requests returned `403`. Sample denial requests completed in 0.41–0.81 seconds, while sampled successful full turns completed in 12.2–38.4 seconds. The denied account remained eligible to already queued and later automatic work while healthy work was still in flight. A request-local retry was not enough to escape it across concurrent or later requests.

Entitlement-aware TeamClaude routing

## Predicate check: account versus model

The quarantine key was checked against the live predicate rather than inferred from the error text. Across 24 fixed proxy observations attributed to failing Anthropic organization `56d37f73`, all 24 returned the identical structured denial across three Claude model families:

| Observed route identity | `claude-fable-5` | `claude-opus-4-8` | `claude-sonnet-5` | `gpt-5.6-sol` |
| --- | --- | --- | --- | --- |
| Failing Anthropic org `56d37f73` | `403 oauth_not_allowed_for_organization` | `403 oauth_not_allowed_for_organization` | `403 oauth_not_allowed_for_organization` | Not routed to this pool |
| Gateway OpenAI/Codex pool | Not routed to this pool | Not routed to this pool | Not routed to this pool | `200`, with no `anthropic-organization-id` |
| Independently pinned healthy Anthropic member | Not independently confirmed | Unfilled | Unfilled | Not applicable |

This rejects a model-scoped interpretation for the failing route. A per-model cooldown would leave it eligible for Opus and Sonnet even though it returns the same denial there. The observation that changing the model can make a request work is explained by a pool change: `gpt-5.6-sol` is served by separate gateway OpenAI/Codex members, not by the denied Anthropic account. One gateway endpoint makes that look like a same-account model swap when it is not.

The healthy-member complement remains explicit rather than inferred. It was not independently confirmed because the fast-fail route kept attracting the sampled Anthropic requests.

## What changes

- Classify only the structured OAuth entitlement denial above. Message text, malformed bodies, other permission errors, and generic `403` responses do not trigger quarantine.
- Exclude that account from automatic routing and exhausted-fleet probes for five minutes. The cooldown is process-local, expires without a restart, extends but never shortens, and is not persisted as quota state.
- Recheck the cooldown after storm-control admission. Requests that selected the account before another request observed the denial rotate before they send upstream.
- Include the cooldown in synthetic `retry-after` calculation when every account is unavailable.
- Show the remaining cooldown in `teamclaude status`.
- Add `x-teamclaude-account` to terminal relayed upstream responses. Its URI-encoded value names the final serving account, overwrites any upstream spoof, handles arbitrary configured display names, and is omitted from synthetic TeamClaude errors.
- When every configured account returns the exact entitlement denial, return a diagnostic that says no account served the request, names the denied accounts and code, and recommends automatic re-admission or a different eligible pin instead of incorrectly recommending `teamclaude login`.

The `403` inspection is capped at 64 KiB so a configurable alternate upstream cannot force an unbounded diagnostic buffer.

## Workaround versus fix

Deterministic pinning already exists. `TC_ACCT=<stable-account-identity> teamclaude run` works in both MITM and base-URL modes, targets exactly one account, and never fails over. The deprecated direct-proxy path form `/tc-acct/<name>/...` was also exercised against a local fork and successfully served from the requested account.

Pinning is the deterministic caller workaround. It does not heal automatic rotation for other callers. The cross-request cooldown is the fleet-routing fix, and `x-teamclaude-account` makes the serving decision observable.

## Verification

- `npm test`: 541 passed, 0 failed
- `npm run lint`: passed
- Real HTTP proxy tests cover exact classification, cross-request quarantine, cooldown expiry, queued-request rotation, full-fleet `retry-after`, honest all-entitlement-denied diagnostics, strict caller pins, generic and non-JSON `403` behavior, oversized error bodies, final-account attribution, upstream spoof replacement, Unicode header safety, and synthetic-error omission.

## Follow-ups kept out of this routing change

Two existing CLI defects are reproducible on `v1.1.13` and are intentionally not mixed into this PR:

1. `teamclaude route --help` exits 1 with `ReferenceError: Cannot access 'ROUTE_USAGE' before initialization`. Top-level command dispatch reaches `routeCommand()` before the later `const ROUTE_USAGE` initializer.
2. `teamclaude accounts` displays one-based numeric indices, but `teamclaude disable 1` exits 1 with `Account "1" not found`. The enable/disable path resolves stable identity strings only. Either the command help should stop promising an index, or the displayed index should be resolved explicitly for this one-shot command.

Operationally, the affected work was routed around through GPT-5.6 Sol. This is a durable routing correction, not an emergency workaround.


MagicalTux (dismissed):
The quarantine itself is correct, bounded (5-minute extend-only cooldown, process-local, lazy expiry, never persisted, so it cannot strand an account) and well tested — happy to merge it. Two things to drop first:

1. **`x-teamclaude-account` on every relayed response.** Account names are `email (Org)`, so on a shared proxy (`proxy.host: 0.0.0.0` + `proxy.apiKey`) every client now learns whose personal account served their request, on every response, with no opt-out. It is also unrelated to the quarantine. Please remove it (with its tests and doc paragraph), or gate it behind a config flag that defaults off — `/teamclaude/status` is the observability surface.
2. **`docs/assets/teamclaude-entitlement-routing.png`** (1.5 MB, not referenced by any doc). Drop it.

Minor: the admit-recheck branch recurses without the `retryCount < maxRetries` guard the other retry sites carry. It terminates via `ctx.tried`, but worth matching.

PR #194, opened 2026-08-26
feat(proxy): attribute usage by configured request headers

### LINKED ISSUE #196: Show project and client on dashboard session usage
The dashboard session usage table should include the project and client associated with each Claude Code session when that attribution is available.

Expected behavior:
- Session usage rows show the session id, project, client, state, request count, token counts, and last-used time.
- The dashboard can filter the session list by project and by client.
- Sessions without attribution still render with an empty/unknown value rather than disappearing.

This depends on TeamClaude already receiving per-request usage dimensions and client attribution.

Frozen666:
Rebased onto master (1.1.14). #186 has merged, so the stale copy of it that used to sit in this branch is gone; what remains is #188 plus this change. Stacked on #188.

## Summary
- add configurable request-header usage dimensions under `proxy.usageDimensions`, fed by `ANTHROPIC_CUSTOM_HEADERS`
- expose dimension counters in `teamclaude status`, `/teamclaude/status`, and the dashboard, alongside the existing `clients` payload
- add a per-session table built on #192's numbers — the tokens each response reported, cache included, per weekly bucket — behind `proxy.sessionDetail` (default off)
- make the dashboard usage tables sortable, and the session table filterable by project or client
- persist configured dimensions in the state file
- document local and CI setup

There is no longer a `session` usage dimension. #192 already meters per-session cost from the response usage; counting `input_tokens` + `output_tokens` a second time would have been the wrong number by orders of magnitude on a cached session.

Closes #196.

## Testing
- `npm run lint`
- `npm test` (884 tests)

Also verified end to end against a stub upstream: a configured dimension header is booked here and absent from the forwarded request, `teamclaude status` grows a `Project usage` section and no session section, and `sessions.items` appears only with `proxy.sessionDetail` on.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -5,7 +5,12 @@
     "clientKeys": [
       { "name": "alice", "key": "tc-alice-own-secret" },
       { "name": "bob", "key": "tc-bob-own-secret" }
-    ]
+    ],
+    "usageDimensions": [
+      { "name": "project", "header": "x-teamclaude-project" },
+      { "name": "ref", "header": "x-teamclaude-ref" }
+    ],
+    "sessionDetail": false
   },
   "upstream": "https://api.anthropic.com",
   "switchThreshold": 0.98,
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -41,6 +41,8 @@ Volatile runtime state (observed quota) is written separately to `teamclaude.sta
 | `proxy.host` | Interface to bind. Defaults to `127.0.0.1` (localhost only). Set to `0.0.0.0` (or override with env `TEAMCLAUDE_HOST`) to accept off-box clients — in which case **set `proxy.apiKey`**, since remote clients must present it (via `x-api-key`, or `Proxy-Authorization` for CONNECT/HTTPS-proxy usage); loopback is always exempt |
 | `proxy.apiKey` | API key clients use to authenticate with the proxy (required for any non-loopback client; the proxy injects real account tokens, so an unauthenticated open port would leak them) |
 | `proxy.clientKeys` | Optional per-client keys: `[{ "name": "alice", "key": "tc-…" }, …]`. Each entry authenticates exactly like `proxy.apiKey`, and the tokens its responses report are booked against `name` — per-client usage shows up under `clients` in `/teamclaude/status`, in `teamclaude status`, and (with `--activity-log`) as a `[name]` prefix on each request line. Counters persist in the state file. Traffic on the shared `proxy.apiKey` or the loopback exemption stays unattributed, so give every consumer their own entry when you want complete stats. Edits apply live via `POST /teamclaude/reload` |
+| `proxy.usageDimensions` | Optional request-header usage dimensions: `[{ "name": "project", "header": "x-teamclaude-project" }, …]`. For each request the proxy reads the configured headers and books the response tokens against their sanitized values, shown under `usageDimensions` in `/teamclaude/status`, `teamclaude status`, and the dashboard. A request that omits a header is simply unattributed for that dimension. Counters persist in the state file, and each dimension is capped at 500 distinct values — further values are summed into `(other)` rather than evicting existing rows. Edits apply live via `POST /teamclaude/reload` |
+| `proxy.sessionDetail` | Adds a per-session breakdown (`sessions.items`) to `/teamclaude/status` and the dashboard: one row per session with its id, client, dimension values, pinned accounts, and the tokens it spent per weekly bucket. **Off by default** — any holder of any proxy key can read status, so on a shared proxy this shows every consumer what every other consumer is working on. The aggregate `sessions` counts are unaffected and always present |
 | `upstream` | Upstream API base URL |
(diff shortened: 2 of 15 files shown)


ak2k:
Flagging an overlap: #192 (stacked #189→#190→#192) works the same problem at a different layer — it records what a request actually cost in tokens, per session and per model, from the API response, where this PR attributes requests by header-derived dimensions and shows them in the dashboard. I read them as complementary (attribution and display here, token cost accounting there), but they will meet in the status output and the config. If a maintainer picks either up, I'll rebase mine around whichever is merged first — same arrangement as #190/#191.

MagicalTux (changes requested):
Blocked on #186 and #188 (this branch is a merge of both plus three commits), and ak2k's overlap note holds: #192 measures per-session cost from the response usage including cache tokens, while this counts only `input_tokens`/`output_tokens` through #186's `onUsage` — for cached sessions that is off by orders of magnitude. The maintainer should pick one model; if this one, build the session column on #192's numbers.

Concrete issues in the own delta:

1. The `session` dimension is not opt-in: it is always pushed and `renderStatus` prints every dimension, so an unconfigured proxy's `teamclaude status` grows a "Sessions usage" section of up to 1000 lines. Gate it behind config and keep it out of the CLI renderer.
2. `if (this.clients.size >= this.maxKeys) this._evictOldest()` evicts persisted project counters: 500 distinct `x-teamclaude-project` values from any key holder wipe lifetime totals, and the 60 s save makes that permanent. Fold overflow into an `(other)` bucket instead.
3. `export()` uses `const out = {}` / `out[name] = …`, and `__proto__` passes `sanitizeUsageDimensionValue`. Use `Object.create(null)` / `Object.fromEntries`.
4. Configured dimension headers are forwarded upstream unchanged (server.js strips only pseudo/hop-by-hop/`x-api-key`/`accept-encoding`); strip them before forwarding, and consider adding `cookie`/`proxy-authorization` to the reserved names.
5. The dashboard tests are source greps (`/function renderSessionUsage/` and the like) and assert no behaviour.
6. `sessions.items` exposes every session's id, client, project values and counts to any holder of any key — cross-client visibility is a policy question for @MagicalTux.

Frozen666:
Thanks — rebased onto master (1.1.14) and reworked. Two of the premises have moved since the review: #186 merged, so the copy of it inside this branch is gone and the block is lifted; and #192 merged, which settles the accounting-model question in its favour. The session column is now built on #192's numbers, as you asked. @ak2k — no rebase needed on your side, this branch follows yours.

Point by point:

**Model overlap.** There is no longer a `session` usage dimension at all. `SessionTracker` already meters what each response reported per weekly bucket, cache included, and `stats()` now exposes that per session instead of the aggregate alone. Counting `input_tokens` + `output_tokens` a second time through `onUsage` would have produced a second, wrong number beside the right one.

**1 — `session` not opt-in / 1000-line status.** Gone rather than gated. `UsageDimensionTracker` now holds only operator-configured dimensions, so `renderStatus` iterates a list bounded by the config file, and each tracker is still key-capped. A proxy with no `usageDimensions` prints nothing new. There is a test asserting that a 300-row `sessions.items` produces no CLI section.

**2 — eviction destroys persisted counters.** Replaced with an `(other)` fold: at the cap, a new value is summed into a reserved `(other)` row and nothing is ever deleted. Tests cover both the live case and a restored snapshot being the least-recently-used row.

**3 — `{}` and `__proto__`.** `Object.create(null)` + `Object.fromEntries` in `UsageDimensionTracker.export()`, matching what you did in `ClientUsageTracker`. Worth noting where the vector actually is: a dimension *name* is operator config and rejected by `DIMENSION_NAME_RE`, but a *value* comes from a request header and `__proto__` passes sanitization — that is the case the test pins.

**4 — headers forwarded upstream.** Configured dimension headers are now stripped in the upstream header builder, next to the hop-by-hop set. Tested over a real socket: the configured header does not reach upstream while an unconfigured `x-teamclaude-*` header still does, so this is not a blanket filter. `cookie` and `proxy-authorization` are added to the reserved names, and the docs no longer claim the headers are forwarded.

**5 — grep tests.** The page's pure logic (row building, filtering, sorting, the quota-bar rows) is now exported from `dashboard.js` and serialized into the page with `toString()`, so the tests exercise the same functions the browser runs. One test asserts the serialization itself, which is the contract that would otherwise break silently if a helper stopped being self-contained. The repo has no DOM dependency and this adds none.

**6 — cross-client visibility.** Config-gated behind `proxy.sessionDetail`, **default off**: with it off, `stats()` has no `items` key and the payload is unchanged from master. Please confirm the default is what you want. Two things I did not decide for you: whether it should instead be scoped per client (each key holder sees only their own sessions), and whether it belongs under `proxy` at all rather than beside the dashboard config. Happy to change either.

One thing outside this PR that you may want to look at separately: the dashboard's quota bars were hard-coding Sonnet and Fable, which `40e44d5` made incomplete — they now read `quota.scopedWeekly`, so a family upstream starts metering shows up without a release. That is in #188, since it is a 1.1.14 correctness fix rather than part of this feature.

PR #198, opened 2026-08-27
fix(server): treat a DNS failure as transient instead of failing over across accounts

lexfrei:
A DNS failure is currently classified as "this account's transport is broken", which is the one thing it never is.

`forwardRequest` sorts upstream failures into transient and account-specific. Transient ones close the connection so the client reconnects cleanly; everything else fails over to the next account. `ENOTFOUND` and `EAI_AGAIN` were in neither list, so they fell into the account-specific branch by default. When the resolver hiccups, one local failure is amplified into one failure per account: the fleet is walked and the client gets a `502`, or the synthetic "all accounts exhausted" `429`, instead of the fast reset that would have let it retry. I hit this on a live proxy; that is where the diff comes from.

This PR carries two commits and absorbs #184, which fixed the same amplification and was posted first; coordinated with its author in the comments below, so a maintainer only has to read one PR.

The first commit is #184's, cherry-picked with authorship preserved. Upstream failures sort into two sets: socket-scoped codes stay unconditionally transient, and host-scoped codes (`ENOTFOUND`, `EAI_AGAIN`, `EHOSTUNREACH`, `ENETUNREACH`, `ENETDOWN`) are transient only while no untried account would dial a different host. Codes are read from the error, its `cause`, and its children, because global fetch puts the real error on `cause` and the happy-eyeballs dialer reports an `AggregateError` with no top-level code. The `fetch failed` message check runs after the code checks, since that message is every failure on the global-fetch path and reading it first would answer for the whole transport before the host-scoped arm ever ran. `EPIPE` joins the socket set as the write-side sibling of `ECONNRESET`. My original version of this fix read only `err.code` and checked the message first, so under the global-fetch transport it never reached the conditional arm at all; that ordering is the better foundation, which is why that commit comes first and mine builds on it.

The second commit tightens what "another host" means. Untried is not enough: account selection gates on routes and the disabled flag, so a different-host account this request can never legally route to gives failover nothing to reach, and one such account would otherwise disarm the fast reset on a routed fleet. The scan reuses the account manager's own eligibility predicate rather than restating route logic — deliberately not the selection entry point itself, which can arm the probe cooldown as a side effect. Eligibility is checked for the executor model only: when no account satisfies both the executor and the advisor model, selection degrades to executor-only routing, so failover reaches every executor-eligible account, and gating the scan on the advisor would call a reachable healthy host "nowhere to go". Hosts are compared by hostname, so a port or path difference is not mistaken for a second host.

Pinned requests are carved out. A pin restricts the request to one account, so there is no failover to skip; keeping a resolution failure on the failover path means the client gets the informative pinned-unavailable `429` instead of a bare connection reset.

Sixteen tests across three files: unit coverage of both directions of every classification (including the wrapped-`AggregateError` shape and the message-read-last ordering), fleet-level tests that count what one dead name costs in attempts, and end-to-end tests for the failover, route-ineligibility, advisor-degradation, and pinned paths. The unresolvable hosts use RFC 2606 `.invalid` names, and the fleet tests assert they actually observed a resolution failure, so a resolver that hijacks unknown names fails loudly.


ak2k:
This overlaps #184, posted 08-20 — same reclassification of `ENOTFOUND`/`EAI_AGAIN` out of the failover branch. Your version goes further than mine on the per-account `upstream` case: the host-scoped guard alone makes it the better fix for routed fleets, and the pin carve-out and eligibility predicate are improvements #184 doesn't have. Happy to coordinate rather than make a maintainer read both: I can narrow #184 to whatever this doesn't cover, or withdraw it in favor of this PR if you'd fold in #184's test coverage of the close path. Either way, flagging it now so neither of us is surprised at review time.

lexfrei:
@ak2k Took you up on this. #198 now carries your commit as-is, cherry-picked with authorship preserved, and my eligibility/pin layer is rebuilt on top of your classification. Merging the two showed your foundation was the right one: my version read only `err.code` and checked the `fetch failed` message first, so under the global-fetch transport a wrapped `ENOTFOUND` never reached the conditional arm. Your code-before-message ordering fixes that, which is why your commit comes first.

One delta from #184 as posted: the other-host scan now goes through the manager's eligibility predicate (executor model only, since selection degrades to executor-only routing), which also closes the disabled-account gap your ECONNREFUSED note described. All your tests are in; the only edit is that comment, reworded to past tense, with your guard test still pinning ECONNREFUSED as unconditional.

If that shape works for you, withdrawing #184 leaves the maintainer one read. Thanks for flagging the overlap before review time.

ak2k:
Looks great — withdrawing #184 in favor of this. Thanks for the clean consolidation.

PR #199, opened 2026-08-27
fix(accounts): pair config entries to accounts by entry id, not by position

### LINKED ISSUE #201: The TUI applies manager indices to the config list in three places
Three places in `src/tui.js` pair the two account lists by position. They are not index-aligned: `resolveAccounts` drops credential-less entries at startup, so from the first drop every later index differs.

- `_doRemove` splices `config.accounts` at the manager index (`src/tui.js:963`), so it can delete the wrong row.
- `_doToggleDisabled` writes `disabled` at the manager index (`:976`), setting it on the wrong row.
- `_doImport` goes the other way. `idx` comes from `findUpsertTarget(this.config.accounts, entry)`, so the config write at `:911` is correct, but the fallback at `:914` applies that config index to `this.am.accounts`. The block below it then writes `accountUuid`, `orgUuid` and `orgName` onto whatever live account it landed on. That is identity corruption rather than a stale field: every later pairing reasons about the wrong account.

Identity is not enough to resolve them, since two entries for the same person are one identity and two rows. An id on the config entry, carried into the account built from it, separates them.

I may get to this if time allows. Filing it so the finding exists as an artifact rather than dying with the session that found it.

lexfrei:
`resolveAccounts` drops every entry without a usable credential, so from the first drop a config index and a manager index name different accounts for the life of the process. Several sites index one list with the other's index. A refreshed token lands on a stranger's config row and is persisted there; the TUI deletes or disables the wrong row; an import writes one person's credential and organization onto another person's live account.

Each config entry now carries an `id`, minted where entries are created and read through into the account built from them. Pairing is a lookup on that id. Where the id is missing on both sides the code pairs nothing rather than guessing.

Seven crossing sites are converted: the token-refresh write and the save in `index.js`, `_doRemove`, `_doToggleDisabled`, `_doImport` and `_doAddKey` in `tui.js`, and the shared upsert path. Two of those were not in the first version of this PR, and one of them points the other way, applying a config index to the manager list.

An earlier version of this PR paired by identity through a 312-line heuristic. That is gone. Identity cannot separate two entries for one person, which is the case the pairing exists for, and the module's own docstring said so: it named two states that break its guarantee and said no arrangement of its passes reaches them. An id reaches both. What replaces the heuristic is 141 lines across two modules, and a 762-line enumeration test becomes 240 ordinary ones, because id equality is transitive and the property no longer has to be enumerated.

Reverting one file at a time shows what each part carries. `tui.js` alone at master fails 6 of 8 crossing tests. Master's positional refresh and save fails 8 of 12 pairing tests. `makeAccount` alone at master, so the id is never carried through, fails 15 across the suite.

`config.json` gains one auto-managed field. Delete it and it is minted again; copy an account block and the duplicate is re-minted on the second entry. `docs/accounts.md` says so.

The disk axis is unchanged: `mergeAccountsForSave` and `findConfigAccount` still resolve by first identity match, which #202 and #203 describe. With an id now on both sides, one change closes both.

33 new tests, 687 pass, up from 654, and the full run is no slower because the enumeration test left with the heuristic.

Closes #201


MagicalTux:
Read in full. The positional crossing is real (the TUI save pairs `config.accounts[i]` with `accountManager.accounts[i]`, and the refresh hook mirrors by manager index), and the targeted tests would fail on master. Two things for @MagicalTux to decide: (a) the module's own docstring says the problem is only fully solvable with a stable id assigned where config entries are created and carried into `makeAccount` — that fix is small and would make the 312-line heuristic unnecessary; (b) the enumeration test is large — please post its measured runtime. Also: this hard-conflicts with #197 (same function), which I'm merging first, so a rebase is needed; and, as the description says, it fixes none of #200–#205.

lexfrei:
Taken, both points. The heuristic is gone and the id design replaces it: 141 lines across two modules instead of 312, and the enumeration test drops from 762 lines to 240 ordinary ones. 800 insertions against the 1118 you read.

Your inference went further than the docstring, which claimed the id was needed for a residual rather than sufficient for the whole thing, so I tested it before deleting anything. It holds. The id closes the two states the docstring calls unreachable, plus two crossing sites the first version never touched. One of those wrote a live account's `accountUuid` and `orgUuid` from a stranger's import, which the old description listed as a known unfixed cost.

The runtime you asked for: 0.81 to 0.86 seconds wall clock across three runs, with 697 ms of it in three enumerating tests. That file no longer exists, so the answer is moot, but you asked and guessing at it would have been worse than measuring.

One correction to #201, which this closes. Its third bullet blamed the config write in `_doImport`. That write is correct: `idx` comes from `findUpsertTarget` over `this.config.accounts`. The crossing is the fallback below it putting that config index into `this.am.accounts`. I have corrected the issue. It also proposes resolving by identity, and identity is what cannot separate two entries of one person, so the fix here is by id instead.

Rebased onto master. #197 had already moved `syncAccountsFromDisk` out of `index.js`, so the largest hunk of the old version had no home and was rewritten against the new shape rather than merged into it.

Two things left alone, both pre-existing and both already filed. The disk axis still resolves by first identity match, which is #202 and #203; with an id on both sides now, one change closes them together. And a re-import onto a credential-less entry still does not bring the account into service until the next reload, which master also did while filling the gap by hijacking a live account instead.

PR #209, opened 2026-08-28
fix(run): bootstrap client auth through proxy

atmigtnca:
## Summary
- bootstrap Claude Code through the proxy when local OAuth is missing or expired
- preserve valid local OAuth and existing `ANTHROPIC_AUTH_TOKEN` flows
- keep `--auto-fallback` semantics unchanged
- shell-quote emitted credentials for eval safety

## Limitations
Remote Control and claude.ai file-transfer passthrough endpoints still require client-side authentication.

## Verification
- `npm test`: 533/533 passed
- `npm run lint`: clean
- isolated HOME end-to-end invocation passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,11 @@ teamclaude run       # in another terminal: Claude Code through the proxy
 
 Already logged into Claude Code? `teamclaude import` takes its credentials instead of a fresh OAuth round. API keys, and one email holding accounts in several orgs, are covered in [docs/accounts.md](docs/accounts.md).
 
+`teamclaude run` does not require a separate Claude Code login for normal API
+requests. When local Claude OAuth is unavailable or expired, proxy credential
+mode supplies a local-only bootstrap key; the proxy replaces that placeholder
+with the selected TeamClaude account credential before forwarding.
+
 ## What it does
 
 - Rotates to the next account when the 5h session or 7d weekly bucket reaches the threshold (98% by default), preferring the account whose weekly quota resets soonest.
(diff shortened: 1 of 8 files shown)


MagicalTux (changes requested):
Two blocking problems, then a premise question for @MagicalTux.

1. **Expiry test is wrong.** `needsProxyClientCredential` treats an expired `accessToken` as "no local OAuth" and ignores `refreshToken`. Claude Code refreshes its own access token on start (the proxy relays `/v1/oauth/token` untouched), and access tokens expire within hours, so the ordinary "haven't run claude since yesterday" case would silently drop the user into API-key mode — the exact outcome `run`/`env` currently guard against. At minimum require `!refreshToken`; arguably only a missing file should qualify.
2. **Unverified upstream compatibility.** Nothing in the proxy adds `anthropic-beta: oauth-2025-04-20` to forwarded `/v1/messages` (the string exists only in the usage probe). In API-key mode Claude Code shapes its requests for key auth; whether upstream accepts an injected OAuth Bearer on such a request is not shown by any test here — they check the emitted env lines and the pure predicate only. Please add a proxied-request test in bootstrap mode, or document a real run in the description.

Also: `CLAUDE_CONFIG_DIR` is not honoured, so a relocated Claude Code config reads as "no OAuth" on every launch; `run` checks only `ANTHROPIC_AUTH_TOKEN` before overwriting a user-supplied `ANTHROPIC_API_KEY`; and every `env`/`run` now reads the credentials file (and on macOS may spawn `security`, which can prompt).

**Premise:** this inverts the documented stance that `run` never emits `ANTHROPIC_API_KEY`, because API-key mode changes Claude Code's own behaviour (and, as your Limitations note says, breaks Remote Control and file transfers). Whether teamclaude should paper over a missing Claude Code login at all is the maintainer's decision, so I'm not merging on review alone.

PR #213, opened 2026-08-30
feat(tui): name activity rows after the Claude Code session

rikbrown:
## What

The activity log identifies each request by six hex characters of its session id. That separates concurrent sessions, but does not say which is which. This PR names each row after the Claude Code session that sent it, resolved from Claude Code's own on-disk data under `~/.claude/projects`:

1. the title `/rename` writes (`<session-id>/custom-title.json`),
2. else a `custom-title` record in the session transcript (the only copy for a session renamed before that file existed),
3. else the `ai-title` record Claude Code generates on its own — so most sessions get a usable name with no rename.

A session with none of these keeps its short id.

- New `sessionTitles` config: `{ enabled, width, projectsDir }`, off by default (like probe, warmup and egress pin), every field applied live on config reload.
- Second commit adds a settings-screen toggle (**g** → Session titles).
- Display only — no selection code reads it.
- Only a UUID-shaped session id is looked up, and control characters are stripped from a title before it is printed or logged; the cache expires alongside the tracker's 1 h known window.

## Performance

Lookups stay off the render path: `get()` returns the cached value and schedules the read, and a missing title is cached like a found one, so a nameless session is not rescanned every frame. Only the tail of a transcript is read — measured across 218 transcripts, the last title record sits within 32.7 KB of EOF at p99, so the read window is 64 KB. A 30-second TTL carries a `/rename` into the log without a restart.

## Relationship to #194

Complements #194 rather than competing with it: that PR attributes sessions through `ANTHROPIC_CUSTOM_HEADERS` the user configures and targets the browser dashboard; this one needs no client-side setup and targets the TUI.

## Caveat

Resolution reads the proxy host's own `~/.claude/projects`. When the proxy runs on a different machine from Claude Code, rows degrade to today's short id (`projectsDir` can point elsewhere, e.g. a mount).

## Testing

- `npm test`: 654 tests on master → 699 on this branch, 0 fail.
- `npm run lint` clean.

https://claude.ai/code/session_0137MebGgi2xzZqHj7jLLMRw


MagicalTux (dismissed):
Thanks — the resolver design (cache off the render path, tail-only read, negative caching) is nice. Three things need to change before this can land, plus a design question for the maintainer.

1. **Client-controlled header in filesystem paths.** `sessionId` is the raw `x-claude-code-session-id` header and goes straight into `join(project, sessionId, 'custom-title.json')` and `join(project, \`${sessionId}.jsonl\`)`. `path.join` normalises `..`, so any client that can reach the proxy can make it open arbitrary readable paths every 30 s and print whatever `customTitle`/`aiTitle` it finds. Validate the id before touching disk (Claude Code session ids are UUIDs; at minimum reject `/`, `\`, `..`, NUL and cap the length). `cleanTitle` should also strip control characters — the title is written into terminal escape territory via `fg(...)` and into the `--activity-log` file.
2. **The settings toggle doesn't persist.** `_toggleSessionTitles` sets `this.config.sessionTitles` and calls `saveConfig`, but the `saveConfig` closure in `src/index.js` only copies accounts / sx / switchThreshold / quotaProbeSeconds / warmupSeconds / routes to disk. The next reload then runs `config.sessionTitles = diskConfig.sessionTitles` and flips it back. The test passes only because it injects a fake `saveConfig`. (eventLogging and blockedModels have the same pre-existing gap; I'm fixing that on master separately, and you'll want `sessionTitles` in that block.)
3. **Unbounded cache.** `this.cache` is never evicted, so every distinct session id lives for the life of the process. Expire alongside the tracker's 1 h known-window.

Smaller: `configure()` ignores `projectsDir`, so the docs' "applied live on reload" is not true for that field; `.slice(0, width).padEnd(width)` is UTF-16 based, so a CJK title misaligns every column after it — use the display-width helpers from #179 once that lands; and the label widens every row from 6 to 18 columns even when nothing is titled.

Design question I'm leaving to @MagicalTux: this makes the proxy enumerate `~/.claude/projects` and read the tail of conversation transcripts by default, on a 30 s cadence per visible session. The project's other features that reach beyond the proxy's own state (probe, warmup, egress pin) default to off; I'd suggest `enabled: false` by default here too.

rikbrown:
Thanks, all addressed in 26d4cb6 on top of a rebase: UUID-only session ids, control characters stripped from titles, `sessionTitles` persisted in `saveConfig` alongside #227's fields, cache expiry on the tracker's 1 h window, `projectsDir` applied by `configure()`, display-width padding via #179's helpers, and `enabled: false` by default. Left the 18-column label as is: it only applies while the feature is on.

PR #214, opened 2026-08-30
fix(tui): keep a fraction of a percent in the switch threshold

rikbrown:
## What

Typing `99.5` into the switch-threshold prompt stored `100`. `_doSetThreshold` rounded the entered percent to a whole number before dividing (`Math.round(pct) / 100`), so the fraction was silently discarded — and the screen then showed the rounded value as if it had been accepted.

- The entry is now quantised to a tenth of a percent (`Math.round(pct * 10) / 1000`), which is exactly what the screen displays, so the stored value is the one shown.
- ←→ steps from the exact percent rather than a rounded one, so 99.5 moves to 98.5 instead of snapping to 99.
- `formatPercent` keeps a tenth when the value has one and is exported, so the settings screen and `teamclaude status` print the same figure.

Reported OAuth utilization arrives on a whole-percent grid, so the quota bars are unchanged; the docs note that a fractional threshold only changes the switching outcome for accounts whose used share is continuous.

## Testing

- `npm test`: 525 tests on master → 535 on this branch (new `test/switch-threshold.test.js`), 0 fail.
- `npm run lint` clean.

https://claude.ai/code/session_0137MebGgi2xzZqHj7jLLMRw


MagicalTux:
Pushed one commit on top: the confirmation line in `_doSetThreshold` still printed `Math.round(v * 100)%`, so entering 99.5 stored 0.995, showed 99.5% on the settings row, but logged "Switch threshold set to 100%" — the symptom the PR describes, in its last remaining place. Now uses `formatPercent(v)`, with an assertion on the log line. Note for the description: `formatPercent` also formats the usage bars in `teamclaude status`, and API-key ratios (tokensRemaining/tokensLimit) are continuous, so those can now print a tenth too — harmless, just not "unaffected". Merging once CI is green.

PR #229, opened 2026-09-02
feat(quota): run the quota probe by default

thomasahle:
The quota probe reads `/api/oauth/usage`, which reports utilization **without consuming message quota**. Being opt-in made sense when it looked like one more way to learn the same numbers the response headers already carry. It isn't — it is the only path to two things.

**The Sonnet weekly bucket has no response header at all.** `unified7dSonnet` is populated exclusively by `applyUsageData`. With the probe off it stays `null` forever, the S7 bar never appears, and Sonnet requests are gated by the shared weekly bucket alone.

**A spent family bucket can only be refreshed by a request of that family.** The `7d_oi` headers ride on Fable responses and appear on no other model's response — verified on a Max account, same account seconds apart:

```
claude-fable-5-1 → 200   anthropic-ratelimit-unified-7d_oi-utilization = 0.0
claude-opus-5    → 200   (no 7d_oi header at all)
```

So once the bucket reads spent, selection stops sending the one thing that could correct it. Observed live on a six-account fleet: five accounts pinned at the switch threshold for Fable with cached resets days out, while `/api/oauth/usage` reported `{"kind":"weekly_scoped","percent":0,"scope":{"model":{"display_name":"Fable"}}}` on those same accounts. Passive-only mode has no way to see that. (#226 adds a staleness escape for that case; this PR removes the need to rely on it.)

## The change

Default to 300s. The distinction that matters is **absent** (take the default) vs. an **explicit 0** (stay off), so `teamclaude probe off` sticks, and any config that already set an interval keeps it.

That distinction is why this isn't a one-line diff: `config.quotaProbeSeconds || 0` appeared in four places and cannot tell "off" from "unset". The default now lives in `config.js` as `DEFAULT_QUOTA_PROBE_SECONDS`, and every reader goes through `quotaProbeSeconds(config)`.

Keep-warm stays opt-in. That's the line this draws: the probe spends nothing, keep-warm spends real quota.

## Cost

One `GET /api/oauth/usage` per OAuth account per interval — on a six-account fleet, 72 requests/hour against an endpoint that exists to be polled. API-key and third-party accounts are skipped. The 30s floor is unchanged.

## Tests

`test/quota-probe-default.test.js` pins the absent/explicit-0/explicit-value cases and that a fresh config writes the default out. Suite 546/546, `eslint` clean.

https://claude.ai/code/session_019FgVD2djKUuKWr4maN34H9

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Step-by-step lifecycle: [docs/routing.md](docs/routing.md#request-lifecycle).
 | [Accounts](docs/accounts.md) | OAuth login, import, API keys, multiple orgs, third-party backends |
 | [Usage](docs/usage.md) | Server and TUI, running Claude Code, shell alias, command reference, logging |
 | [Routing](docs/routing.md) | Rotation, the two kinds of 429, storm control, model routes, session spreading, pinning, prompt cache |
-| [Quota](docs/quota.md) | Quota probe, keep-warm, holding on exhaustion |
+| [Quota](docs/quota.md) | Quota probe (on by default), keep-warm, holding on exhaustion |
 | [Configuration](docs/configuration.md) | Config format, every field, environment variables, network tuning |
 | [Proxy modes](docs/proxy-modes.md) | MITM forward proxy, sx.org residential egress |
 | [Compliance](docs/compliance.md) | Terms of service notes |
--- a/docs/quota.md
+++ b/docs/quota.md
@@ -10,17 +10,17 @@ Observed quota is persisted to `teamclaude.state.json` next to the config, so ro
 
 ## Quota probe
 
-If you'd rather keep idle accounts' quota fresh, enable the background probe:
+The background probe keeps idle accounts' quota fresh. It is **on by default**, every 300s:
 
 ```bash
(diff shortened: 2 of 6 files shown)


MagicalTux:
Thanks for the write-up — the two facts it rests on (no Sonnet header; `7d_oi` only rides on Fable responses) are right, and #226 landed for the second one.

Closing this one on policy, though: teamclaude stays passive by default. The probe is zero-spend, but it is still the proxy originating requests to Anthropic on its own schedule from every configured account, and that is the line the project draws for defaults (same as keep-warm, the egress pin and session titles). Every active feature is opt-in and off until the operator asks.

The opt-in is already there and does what this PR wants for anyone who turns it on: `quotaProbeSeconds` in the config, `teamclaude probe 300`, or the **Quota probe** row on the TUI settings screen — applied live on reload. If the gap is discoverability, a hint would be welcome: e.g. when the TUI shows a `⊘ Sonnet`/`⊘ Fable` tag (or `unified7dSonnet` has never been seen) with the probe off, mention that `teamclaude probe 300` reads those buckets at zero spend. Happy to take that as a small PR.

Issue #72, opened 2026-07-04
Thinking of a new name


MagicalTux:
This project was originally named TeamClaude, but considering Anthropic's general attitude toward protecting what is [a fairly common first name](https://www.behindthename.com/name/claude/top/france), and the possibility TeamClaude could be useful for other use cases, I'm considering renaming the project and open to ideas.

Main concepts include routing and transparent proxying.

teocns:
I think adopting some non-brandable name like `claude-multilogin` won't be an issue given there are thousands of other `claude-<whatever>` repositories.

roderik:
Or add codex support and make it teamagent?  <- hidden feature request ;)

jfarre20:
CARL - Code Agent Routing Layer

MagicalTux:
I like this a lot. CARL.

vanseed:
Claude-LB，Simple and clear

teocns:
Another brilliant one `claude-multiauth` since `multilogin` can conflict with https://multilogin.com/

nova28:
Hi @MagicalTux, with the new #246 not sure you should be looking into the project name again and/or extend the support to other vendors.

ak2k:
As another suggestion, perhaps `team-router`.

Two things to know before picking it. The GitHub org [isn't available](https://github.com/TeamRouter/), and it belongs to a product that already uses the name (it has a user-feedback repo and teamrouter.com); a 2025 repo, [eeshhwaq/TeamRouter](https://github.com/eeshhwaq/TeamRouter), is also an LLM proxy for teams. On the other hand the npm names `team-router` and `teamrouter` are unpublished, and I assume you'd keep the repo on KarpelesLab, where the scope covers it.

PR #246, opened 2026-09-03
feat(codex): pool Codex subscriptions alongside Claude accounts

nova28:
Adds a `provider` seam and a second provider, `codex`, so an OpenAI Codex subscription can be pooled and rotated the same way Claude accounts are.

This is a nod at #72 — "the possibility TeamClaude could be useful for other use cases" — and at the hidden feature request in that thread. It also puts a boundary on how far that generalisation reaches; see the last section.

## What it is not

A passthrough, not a translation layer. Each provider forwards its own protocol untouched; only the injected credential and the destination change. Converting between protocols would mean re-serialising tool calls, streaming events and cache breakpoints, which is the fidelity loss this proxy exists to avoid.

## One port, path-routed

The request path says which pool of accounts is eligible: Claude Code posts to `/v1/messages`, Codex to `/backend-api/codex/responses`. No second listener, no client-supplied hint that could disagree with the body — one port, one config, one TUI.

A provider partition is absolute (an Anthropic account cannot serve an OpenAI Responses request), so it is expressed by **widening the per-request exclude set** rather than threading a parameter through the rotation logic. Everything downstream already reads `exclude`, so cursors, pinning, session affinity, probing and preemption keep working unchanged, and a config with no Codex accounts produces the identical set it did before.

## Notes from building it

- **The credential is a bearer token plus a `ChatGPT-Account-Id` header.** That header is OpenAI's counterpart to the `account_uuid` this proxy patches into an Anthropic request body — so the Codex path does no body rewrite at all. The Anthropic-only rewrites are now guarded: a Responses body has no `metadata.user_id` to patch and no Anthropic tool-pairing rule to repair.
- **The upstream is `chatgpt.com`, not `api.openai.com`.** A subscription token is a ChatGPT token; the API platform refuses it with `Missing scopes: api.responses.write`. Codex appends `/responses` and `/models` to its configured `base_url`, so pointing it at `<proxy>/backend-api/codex` makes it emit exactly the paths the ChatGPT backend expects, forwarded verbatim.
- **`OPENAI_BASE_URL` does not work** for redirecting a ChatGPT-authenticated Codex — it is ignored. `model_providers` in `~/.codex/config.toml` is the supported route. Both were tested.
- **Tokens refresh** against `auth.openai.com` with the Codex CLI's own client id, returning the same `{ accessToken, refreshToken, expiresAt }` shape as the Anthropic path.
- **`makeAccount` is an allow-list**, so `provider` and `accountId` are carried through it explicitly. A test caught this — without it the fields were silently dropped and selection crossed providers.

## Backwards compatibility

An account with no `provider` is Anthropic. The configured `upstream` stays the Anthropic default rather than becoming global, since existing configs set it meaning "where Anthropic lives" — applying it to Codex would silently send OpenAI traffic to an Anthropic host.

## Scope

Quota-driven rotation stays Anthropic-only: the switch threshold, reset countdowns and the TUI's bars read `anthropic-ratelimit-unified-*`. Codex accounts rotate on upstream rejection, as a third-party backend does today. Codex reports its own `primary`/`secondary` windows with `resets_at`, so wiring those in is a natural follow-up rather than something this PR guesses at.

## On how far this generalises (re #72)

Worth recording as a boundary for a routing-layer rename: **not every agent CLI can be proxied.** Google's Antigravity CLI (`agy`) has no base-URL flag, ignores `HTTPS_PROXY` (google-antigravity/antigravity-cli#72), and does its own network I/O from a single binary — there is no seam to attach to at all. Codex fits this architecture because it exposes one; that is the property worth selecting for.

## Testing

- 882 tests pass, lint clean.
- New: `test/provider.test.js`, `test/provider-routing.test.js`, `test/codex-auth.test.js`.
- Verified end to end against a live Codex subscription — `codex exec` through the proxy returns a real completion with tokens billed to the pooled account, and the Anthropic path is unchanged.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -28,6 +28,12 @@
       "refreshToken": "sk-ant-ort01-your-refresh-token"
     },
     {
+      "name": "codex@example.com",
+      "type": "oauth",
+      "provider": "codex",
+      "priority": 0
+    },
+    {
       "name": "api-fallback",
       "type": "apikey",
       "priority": 10,
--- a/docs/accounts.md
+++ b/docs/accounts.md
@@ -87,6 +87,66 @@ Every account entry carries an `id`, added the first time the config is read and
 
 Hand edits are fine. Leave the `id` alone and it keeps working; delete it and a new one is issued on the next read. If you copy an account block to make a second entry, the duplicated `id` is spotted on the next read and the later of the two gets a fresh one.
 
+## Codex accounts (experimental)
+
+An OpenAI Codex subscription can be pooled alongside your Claude accounts. Add
+an account with `"provider": "codex"` and TeamClaude reads the login the Codex
+CLI already stores in `~/.codex/auth.json`:
+
+```json
+{ "name": "me@example.com", "type": "oauth", "provider": "codex" }
+```
+
+Point another `importFrom` at a different file to pool several Codex logins.
+
+Then tell Codex to reach TeamClaude instead of OpenAI, in `~/.codex/config.toml`:
+
+```toml
+model_provider = "teamclaude"
+
+[model_providers.teamclaude]
+name = "teamclaude"
+base_url = "http://127.0.0.1:3456/backend-api/codex"
+wire_api = "responses"
+```
+
+`OPENAI_BASE_URL` does **not** work for this — a ChatGPT-authenticated Codex
+ignores it. `model_providers` is the supported redirect.
+
+The `/backend-api/codex` suffix matters. A Codex subscription authenticates
+against the ChatGPT backend, not the OpenAI API platform — pointed at
+`api.openai.com` the same token is refused with `Missing scopes:
+api.responses.write`. Codex appends `/responses` and `/models` to `base_url`,
+so this suffix makes it emit exactly the paths the ChatGPT backend expects and
+TeamClaude forwards them verbatim.
+
+### How it shares the port with Claude
+
+One listener serves both CLIs, because the request path says which pool of
+accounts is eligible: Claude Code posts to `/v1/messages`, Codex posts to
+`/backend-api/codex/responses`. An Anthropic account is never offered a Codex
+request and vice versa, so the two rotate independently on one port, one config
+and one TUI.
+
+### What differs from a Claude account
+
+- The credential is injected as `Authorization: Bearer`, plus a
+  `ChatGPT-Account-Id` header. That header is OpenAI's counterpart to the
+  `account_uuid` TeamClaude patches into an Anthropic request body — so the
+  Codex path performs no body rewrite at all.
+- Tokens refresh against `auth.openai.com` using the Codex CLI's own client id.
+- The request body is forwarded untouched. This is a passthrough, not a
+  translation layer: TeamClaude never converts between the Anthropic and OpenAI
+  protocols.
+
+### Current limits
+
+Codex accounts rotate on upstream rejection, the same way a third-party backend
+does. The quota-driven parts of rotation — the switch threshold, reset
+countdowns and the TUI's quota bars — are Anthropic-only for now, because they
+read `anthropic-ratelimit-unified-*` response headers. Codex reports its own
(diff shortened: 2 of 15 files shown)


nova28:
Pushed a second commit that makes the Codex path operable rather than just routable: **quota** and **login**.

## Quota

Codex reports its limits on every response, so those are now normalised into the same fields the Anthropic path fills. `switchThreshold`, reset countdowns and the TUI bars work for Codex accounts with no new UI, and rotation happens *before* upstream refuses instead of after a 429.

Two properties of the header set drove the parser, both taken from a captured live response rather than assumed:

- **Limits arrive in families.** The unnamed family is the account-wide limit; a family carrying `-limit-name` is model-scoped — the counterpart of the `7d_oi` Fable bucket.
- **`primary`/`secondary` are positions, not durations.** The account-wide family put its **7-day** window in `primary`, while the model-scoped family put a **5-hour** window there. Windows are classified by their stated `window-minutes`; reading position would have filed a weekly reading as a session one and rotated on the wrong number. A zero-length window is how this API says "not applicable" and is dropped rather than read as full headroom.

## Login

`teamclaude login --codex` — OAuth against `auth.openai.com` with PKCE. The callback must bind port **1455**, because OpenAI registered exactly one redirect for this client, so a bind clash is reported as that rather than as a generic failure. A Codex `id_token` already carries the email and ChatGPT account id, so unlike the Anthropic path there is no profile call to make.

`CODEX_HOME` is documented as the browserless alternative for pooling several logins, since the Codex CLI isolates credentials by it.

## Verification, stated precisely

- Quota parser tested against the captured header set; threshold handover between two Codex accounts covered by a test.
- Token endpoint and client id confirmed **live** via the refresh grant.
- **The browser login round-trip is not verified end to end.** `auth.openai.com` sits behind a bot check, so the authorize URL cannot be exercised without a real sign-in. Its parameters are extracted from the Codex binary and asserted in tests, but I'd rather flag that than imply I proved it.

899 tests pass, lint clean.

nova28:
Ready for review. Rebased into three focused commits, and the login flow is now **verified end to end** — a real browser sign-in through `teamclaude login --codex` adds the account and writes it to config.

## Commits

1. **`feat(codex): pool Codex subscriptions alongside Claude accounts`** — the `provider` seam, path routing, credential injection, account selection.
2. **`feat(codex): read Codex quota and add a Codex login`** — quota parsing into the existing fields, plus the OAuth login.
3. **`fix(config): apply the proxy setting on a first run`** — an unrelated pre-existing bug, deliberately kept separate; happy to split it into its own PR if you'd prefer.

## Three things testing corrected that reading could not

- **`scope=api` is rejected.** A strings dump of the CLI suggests `openid profile email offline_access api`, but that is two constants running together — OpenAI answers `invalid_scope`, "this client is not allowed to request scope 'api'". The working set is `openid profile email offline_access`, which is consistent with the token being a ChatGPT credential rather than an API-platform one.
- **`primary` is not a duration.** The account-wide family carries its **7-day** window in `primary`; a model-scoped family carries a **5-hour** one there. Classifying by position rather than by `window-minutes` would file a weekly reading as a session one and rotate on the wrong number.
- **First-run self-proxy loop** (commit 3). `loadConfig` publishes the resolved upstream proxy but returns early on ENOENT, so when `loadOrCreateConfig` creates the file nothing installs the setting. The fallback then resolves against an empty config, which has no listener for `isSelfProxy` to compare against — so an operator whose `HTTPS_PROXY` points at their own TeamClaude gets a CONNECT back into the proxy and a 30s timeout, on the one run with no config file to explain it. It applies to `teamclaude login` too, not just the Codex path.

## Verified

- Live `codex exec` through the proxy returns a real completion, billed to the pooled account.
- Live browser sign-in adds and persists the account.
- Token endpoint exercised live via the refresh grant.
- Quota parser tested against a captured header set; threshold handover between two Codex accounts covered.
- 901 tests pass, lint clean. No behaviour change for a config without Codex accounts — an account with no `provider` is Anthropic.

Issue #166, opened 2026-08-10
`unifiedStatus` is stored but not used in account availability decisions


herrozim-ship-it:
## Environment

- teamclaude version observed: 1.1.9
- Confirmed code path still present in: 1.1.12
- Deployment style: local proxy in front of Claude Code over OAuth-backed Max accounts
- Confirmation source: local runtime logs plus line-level inspection recorded in the attached handoff documents

## Observed behavior

During investigation of repeated Fable routing failures, the proxy state showed upstream status values such as:

- Account A: `unifiedStatus: allowed`
- Account B: `unifiedStatus: allowed`
- Account C: `unifiedStatus: allowed_warning`

At the same time, local routing still rejected the affected model family based on local threshold logic.

The handoff notes point to:

- `account-manager.js:1022` storing `unifiedStatus`
- `account-manager.js:761` clearing it
- no matching availability gate that actually uses that value when deciding whether an account can serve a request

The practical effect in the observed run was:

1. upstream had not marked the accounts as rejected
2. local family-bucket utilization hit the configured threshold
3. the proxy treated those accounts as unavailable anyway
4. the user got a local synthetic 429 path instead of behavior that reflected the upstream allow/reject signal

This was especially confusing because the upstream signal and the local decision were pointing in different directions.

## Expected behavior

If teamclaude records an upstream availability signal such as `allowed`, `allowed_warning`, or `rejected`, that signal should influence availability decisions in a clear and deterministic way.

At minimum, I would expect one of these behaviors:

- `rejected` should be able to override optimistic local assumptions
- `allowed` should prevent a purely local stale-state lockout from becoming the only decision input
- if local and upstream signals intentionally diverge, the code path should surface that explicitly in logs and user-facing messaging

## Suggested fix

One possible direction:

1. define the precedence between `unifiedStatus` and local threshold-based family availability
2. apply that precedence in the main availability decision path rather than only storing the upstream field
3. log when a request is blocked by local policy while upstream still says `allowed`
4. expose that distinction in status output so operators can tell "local threshold block" from "upstream rejected"

I am not attached to a specific precedence rule, but today the recorded upstream signal seems effectively disconnected from routing.

## Reproduction

Best-effort reproduction from the observed environment:

1. use multiple OAuth-backed accounts with model-family utilization tracked in state
2. configure `switchThreshold` low enough that a family bucket reaches the local cut line
3. keep upstream account status in `allowed` / `allowed_warning`
4. trigger a request for the affected family
[...]

PR #247, opened 2026-09-03
feat(quota): per-account usage caps

aramalipoor:
`switchThreshold` answers "when should the fleet prefer another account". It cannot answer "this account may spend only 60% of its weekly", for two reasons: it is fleet-wide, and it is a preference rather than a limit — when every account is over it, `_selectProbe` sends one request anyway, on purpose, because a threshold decision can rest on a stale reading and refusing forever is worse than revalidating.

That is right for a rotation threshold and wrong for a budget. A cap the probe path can step over is not a cap.

image

## The setting

```json
{
  "name": "spare@example.com",
  "maxUsage": { "unified5h": 0.6, "unified7d": 0.6, "unified7dFable": 0.8 }
}
```

Same two shapes as `switchThreshold`. A bare number caps every bucket. In a table, `default` covers the buckets that are not listed, and a bucket covered by neither is **uncapped** — a cap is only ever what was asked for, so adding one bucket never quietly constrains the rest.

Keys are the quota field names, which makes a cap model-scoped exactly like a threshold: `unified5h` and `unified7d` stop every model, `unified7dFable` stops Fable alone. The example keeps serving Opus and Sonnet from that account after its Fable budget is gone.

## What "capped" means

At the cap the account receives nothing:

- `unavailableReason` returns `capped`, or `advisor-capped` when it is the advisor model's family that is over. It is checked above every transient state, because a cap is a decision, not an estimate — and status names it, as #166 asked of every other local refusal.
- `_selectProbe` skips it. This is the point of the feature.
- a pinned request (`TC_ACCT`, `/tc-acct/<name>`) is answered with the exhausted response rather than spending past the cap. A pinned request never enters the selection walk, so a budget that only rotation enforced would be one `TC_ACCT` away from being ignored. The pin still never leaks to another account, and nothing changes for an uncapped one.

## Why a new key rather than a per-account `switchThreshold`

Both would compare a bucket to a number, but they mean opposite things about the probe path, and one key that hardens depending on where it is written is worse than two keys that each say what they do. `switchThreshold` keeps its current meaning everywhere.

## Details

- `capExceeded` clears expired windows first, so an account whose window has already reset is never capped on the reading that reset invalidated.
- An account with no `maxUsage` takes an early return; the uncapped path is unchanged, which the "behaves exactly as before" test pins.
- Edits apply live on reload, mirrored onto the memConfig entry like `upstream`/`modelMap` so a later TUI save cannot revert a hand edit.
- `getStatus` reports `maxUsage` per account.

13 tests in `test/account-max-usage.test.js`: resolution (number, table, `default`, unlisted), model scoping, the `>=` boundary, the reset-window case, rotation skipping, the probe path not overriding, an advisor model's family, an uncapped fleet, status output, and an end-to-end pinned request proving no upstream call is made.

`npm test` 871 pass, `npm run lint` clean.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,6 +64,7 @@ Volatile runtime state (observed quota) is written separately to `teamclaude.sta
 | `accounts[].accountUuid` | Anthropic account (person) id; set automatically from the OAuth profile |
 | `accounts[].orgUuid` / `orgName` | Organization the account is scoped to — lets one email hold multiple org accounts |
 | `accounts[].priority` | Rotation preference, lower = preferred (default 0) |
+| `accounts[].maxUsage` | Hard per-account usage cap: a number, or a per-bucket table like `{ "unified7d": 0.6, "unified7dFable": 0.8 }` (same keys as `switchThreshold`; `default` covers unlisted buckets, anything else is uncapped). At the cap the account receives **no** requests — rotation skips it, the all-exhausted revalidation probe skips it, and a pin gets the exhausted answer. Model-scoped like the thresholds, applied live on reload. See [Per-account usage caps](quota.md#per-account-usage-caps) |
 | `accounts[].disabled` | If `true`, the account is excluded from rotation until re-enabled |
 | `accounts[].upstream` | Alternative upstream base URL for this account (e.g. `https://api.deepseek.com/anthropic`). Overrides the global `upstream` for this account only — see [third-party backends](accounts.md#third-party-backend-accounts) |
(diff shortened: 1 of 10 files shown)


aramalipoor:
Force-pushed one correction before review (the branch was minutes old).

`capExceeded` checked only the weekly bucket that *governs* the request, mirroring `_isNearQuota`. That leaks: family spend meters into the shared weekly bucket as well (#175), so a 60% `unified7d` cap did not stop a Fable request whose own bucket was under its 80% cap — and with the family bucket unreported, `_governingWeekly` handed back the shared *value* to be compared against the *family* cap.

The shared buckets now cap every request and a family bucket caps its own family, both applying. That is what the description and the docs already said. Three more tests cover it, including the unreported-bucket case. 873 pass.

aramalipoor:
Not sure if this is a good idea or better to merge it with switchThreshold to 1 flag with switch threshhold + enforce threshold or something

PR #252, opened 2026-09-03
Add timezone-aware rolling warmup schedules

DZamataev:
Supersedes #249 after renaming the fork branch.

## Summary

- add `teamclaude warmup reset HH:MM --timezone AREA/LOCATION` for a daily target reset
- add `teamclaude warmup rolling HH:MM --timezone AREA/LOCATION` for an absolute five-hour cadence anchored to a requested local reset
- persist schedules across restarts, skip missed slots without catch-up traffic, and include timezone-aware confirmation output
- defer an account when Anthropic reports a reset within two minutes of a rolling slot, then retry ten seconds after reset without shifting the global cadence
- expose per-account and tier-weighted fleet quota plus configured warmup metadata at `GET /teamclaude/quota`
- document that warmup sends a minimal real request and therefore consumes a small amount of quota

Rolling warmup is best effort. Ineligible accounts and accounts with an already-running five-hour window are skipped. Delayed retries recheck eligibility and are discarded after the following minute.

## Verification

- `npm run lint`


DZamataev:
Local CI parity update:

PR #253, opened 2026-09-03
Add read-only terminal status dashboard

DZamataev:
Supersedes #251 after renaming the fork branch.

## Summary

- add `teamclaude watch` as a compact read-only terminal dashboard
- show proxy state, Anthropic reachability, account quota bars, session activity, and recent failures
- update the screen in place without terminal flicker while preserving ANSI colors
- support one-shot and JSON output for scripts and non-interactive terminals
- share abort handling with the remote TUI and document the command as part of the normal CLI

The dashboard reads existing local TeamClaude status endpoints and does not alter routing or account state.

## Verification

- `npm run lint`

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ Already logged into Claude Code? `teamclaude import` takes its credentials inste
 ```bash
 teamclaude accounts          # accounts with tier and token status
 teamclaude status            # live proxy status, needs a running server
+teamclaude watch             # read-only terminal dashboard, refreshes every minute
 teamclaude disable <name>    # pause an account without removing it
 teamclaude priority <name> 1 # rotation order, lower = preferred
 teamclaude alias --install   # make plain `claude` go through the proxy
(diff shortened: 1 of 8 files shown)


DZamataev:
Local CI parity update:

MagicalTux:
Closing this one as redundant, and I want to lay out the comparison rather than just say "we have that already", because you could not reasonably have known where the line was.

The screen you have built is good — quota bars, reachability, session activity, recent failures, in-place redraw without flicker, and a non-interactive mode. The trouble is that TeamClaude now has three ways to look at the same data, and this would be a fourth:

- `teamclaude status` — one-shot, plus `--json` for scripts
- `teamclaude attach` — a live dashboard against an already-running server, landed earlier today in #170, which is the read-only-plus-two-keys view for exactly the background-service case
- the TUI itself, when the proxy runs in the foreground

`attach` overlaps this almost completely. It polls the same status endpoint, renders the same fleet picture, and reaches the same deployments. What `watch` adds over it is the flicker-free redraw and a one-shot/JSON path — and `status --json` already covers the second.

Four views of one status payload is a maintenance problem rather than a feature: every new field has to be taught to all of them, and they drift, which is how #172 happened (`status` reported blocked models as available because it did not know about the blocklist that `server` enforced).

Two things from here that I would genuinely take:

- **The flicker-free redraw.** If `attach` repaints worse than this does, that is a real defect in `attach` and a PR against it would be welcome. I fixed a related problem in #173 (the TUI repainted the whole screen twice a second while idle), so that path has known room in it.
- **Anything `attach` shows that it should.** "Anthropic reachability" and "recent failures" are not in it as far as I know. Those are worth adding to the view that already exists.

Sorry to close work that runs. If you want to pick up either of those, I will review it promptly.

PR #254, opened 2026-09-03
Add integrated Git deployment

DZamataev:
Supersedes #250 after renaming the fork branch.

## Summary

- add `teamclaude deploy install/ref/restart/logs/rollback/status/uninstall`
- install immutable Git releases behind atomic `current` and `previous` links
- manage boot-persistent systemd services on Linux and launchd agents on macOS
- validate refs, repository URLs, ownership markers, configuration placement, rollback health, and npm restoration before destructive transitions
- document installation, daily operation, rollback, uninstall, and migration

The deployment keeps account configuration and runtime state outside the release root. Candidate releases must pass the full test suite before activation, and a failed activation restores and verifies the previous release.

## Verification

- `npm run lint`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@ node_modules/
 config.json
 *.log
 package-lock.json
+.env.local
+.planning/
+.worktrees/
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ Step-by-step lifecycle: [docs/routing.md](docs/routing.md#request-lifecycle).
 | [Routing](docs/routing.md) | Rotation, the two kinds of 429, storm control, model routes, session spreading, pinning, prompt cache |
 | [Quota](docs/quota.md) | Quota probe, keep-warm, holding on exhaustion |
 | [Configuration](docs/configuration.md) | Config format, every field, environment variables, network tuning |
+| [Git deployment](docs/deployment.md) | Release worktrees, branch switching, systemd, rollback, and operations |
 | [Proxy modes](docs/proxy-modes.md) | MITM forward proxy, sx.org residential egress |
 | [Compliance](docs/compliance.md) | Terms of service notes |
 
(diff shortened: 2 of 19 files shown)


DZamataev:
Local CI parity update:

MagicalTux:
Thank you for this, and I want to be specific about why it is a no rather than leave you guessing.

The work itself is not the problem. Immutable releases behind atomic `current`/`previous` links, gating activation on the test suite, keeping account config and runtime state outside the release root, validating ownership markers and rollback health before a destructive transition — that is a careful design, and the care shows.

It is a scope decision. TeamClaude is a multi-account Claude proxy: it holds credentials, meters quota, and picks an account per request. `teamclaude deploy` is a release-management system, and a capable one — which is exactly what makes it its own product rather than a feature of this one. Merging it would mean this project owns a deployment tool: its rollback semantics, its interaction with every distro's systemd, its failure modes at 3am on somebody else's server. That is a maintenance surface I am not willing to take on here, independent of how well this particular implementation is written.

It also overlaps what already exists, which is a fair signal that the need is partly met:

- `teamclaude service install` writes the LaunchAgent / systemd `--user` unit and manages the service
- the background self-updater checks once a day and applies on next launch (`autoUpdate`, `TEAMCLAUDE_DISABLE_AUTOUPDATE=1`)
- `npm install -g @karpeleslab/teamclaude` is the install and upgrade path, with npm handling versions and rollback via `@version`

Between those, the gap this closes is "pinned Git refs and instant rollback for people who deploy from source" — real, but narrower than 4,000 lines of new surface here, and well served by a separate tool that calls TeamClaude rather than living inside it.

So: closing, with no reflection on the quality. If you want to publish it standalone I am happy to link it from the docs as a community deployment tool — that gets it to the people who want it without this repo owning it. And if there is a specific thing `service install` cannot do that you needed (a pinned ref, a health-gated restart), open an issue for just that piece and I will look at it on its own merits.

Issue #276, opened 2026-09-06
A family-scoped ineligibility moves the global current account, pulling other families off it too


Frozen666:
`_select` scopes availability per model — its own comment: *"an account whose Fable weekly bucket is spent is still fully usable for other models, so it is only excluded when THIS request targets Fable."* But when that exclusion happens, `_selectNext` writes the replacement into the **global** `currentIndex`, and every other family follows it on their next request.

## Observed

Account A: weekly resets in ~4.5h, Fable bucket at 98%, shared/Opus bucket at 60%, 5h window at ~40% — the account you most want Opus and Sonnet traffic on, precisely because its weekly is about to lapse with 40% unspent. Account B: weekly resets in ~57h.

Current = A. Traffic is Opus and Sonnet on A, as intended. Then one Fable request arrives:

```
12:51:42  Switched to account "B"
12:51:43  POST /v1/messages (claude-opus-5)    → A     ← in flight before the switch
12:51:52  POST /v1/messages (claude-fable-5-1) → B     ← the request that triggered it; correct for Fable
12:52:00  POST /v1/messages (claude-opus-5)    → B     ← wrong: A was fully eligible for Opus
```

From then on Opus and Sonnet burn B's fresh weekly while A's 40% expires unused. This is the case #195's per-route cursor was meant to address: `_selectNext` does read `_previousCursor(model)` — but only to decide whether to *log* "Switched"; the assignment that matters is still `this.currentIndex = best.index`, and `_select` consults `this.accounts[this.currentIndex]` first for every model.

## The mechanism

```js
// _select
const current = this.accounts[this.currentIndex];
if (this._isAvailable(current, model, advisorModel) && !exclude?.has(current.index)) { ... return current; }
const next = this._selectNext(exclude, model, advisorModel);

// _selectNext
const best = this._pickBestAvailable(exclude, model, advisorModel);
if (best) {
  const previous = this._previousCursor(model);           // per-route, from #195
  const switched = previous != null && previous !== best.index;
  this.currentIndex = best.index;                          // global
```

`_isAvailable(current, model)` correctly says "A cannot serve Fable". `_selectNext` correctly picks B for Fable. Then the global cursor moves, so the next Opus request finds `current = B`, available, and stays there. The family-scoped decision has been promoted to a fleet-wide one.

## Suggested fix

[...]

PR #282, opened 2026-09-06
feat(routing): expiry-pressure account selection (rebase of #190)

MagicalTux:
Squashed rebase of #190 onto current master. Original work and evidence are ak2k's; this carries it forward rather than waiting on a rebase.

Squashed rather than replayed because the PR squash-merges anyway, and 23 commits through an API-shape collision on the hot path is 23 chances to resolve it differently each time.

## Two resolutions that would otherwise have silently undone earlier fixes

**`_governingWeekly` is deliberately NOT `_governingWindow().utilization`.** The branch collapsed them, on the reasonable-sounding grounds that "the gate and the ranking cannot answer this question differently". They must:

- the **gate** takes the higher of the family bucket and the shared weekly (#175, merged as #191) — family spend meters into both, so reading the family bucket alone lets an account at `unified7d` 1.00 with `unified7dFable` 0.20 keep serving Fable and push the shared bucket further past its cap on every request;
- the **ratio** must take value and clock from **one** bucket, as this PR's own comment on `_governingWeeklyReset` insists — pairing one bucket's headroom with another's clock prices an account on quota it will lose before it can spend.

`_governingWindow` returns the family bucket alone for a family model (`if (this._weeklyBucketFor(model) !== 'unified7d') return flat;`), so delegating the gate to it reverts #175 exactly. Kept separate, with the reason written where the next reader will look.

**The cursor move now goes through `_setCurrent`, but still only when the exclusion is not model-scoped** — preserving #276's family diversion (merged as #277) while keeping the rollover/ramp bookkeeping the expiry machinery reads.

Also kept: `sweepExpiredQuotas` (#265/#275) alongside the model-aware `refreshExpiredQuotas` signature, and the provider partition and per-provider cursors (#246/#281).

## On the verification

I said earlier I would not hand-merge this because its central claim — flag-off is byte-identical — rests on a differential harness I cannot re-run. That harness is external, so that remains true and I want it stated plainly rather than implied away.

What *does* verify the rebase: the flag-off assertions are committed (`with the knob off the same fleet still rotates on the reset timestamp alone`, `the band is inert with the knob off`), and the entire existing suite runs with the flag off — which is itself a broad flag-off differential across every routing test this repo has. If CI is green, the flag-off path behaves as master does on everything the suite covers.

@ak2k — if you still have the harness, re-running it against this branch would close the remaining gap, and I would welcome the result either way.


MagicalTux:
Rebase status: **5 of 7 failures resolved, 2 remain — and they are a real interaction, not a stale fixture.**

Resolved (both collisions with work that landed while this was open, recorded in the commits):
- the `a rate-limit 429` arm of `SAME_ACCOUNT_RETRIES` rested on the server retrying the same account "never rotating", which #271 changed on purpose;
- `a draining session whose pin is not the current account` passed a fifth argument that was already vestigial on this branch (its own `getActiveAccount` takes four), and #246 has since claimed that slot for `provider` — an object arriving there excludes every subscription account.

## The two that remain

`a chain answers the roll that started it`, both paths. The chain ends on `a` where the test expects `b`, and the mechanism is:

1. `a`'s window rolls, so expiry routing preempts traffic off it — correctly, since a rolled window is now the furthest-dated and should be spent last.
2. The request lands on `b`, which answers `429`.
3. #271's bounded failover hop excludes `ctx.tried` plus the current account. **`a` is in neither** — selection moved off it before any request was sent, so it was never *tried* upstream.
4. The hop therefore returns `a`, and the request settles on the account the rollover had just moved it off, inside the same request.

So the test's own assertion — "the chain settled back onto the account its rollover pushed it off" — is describing a real defect rather than an obsolete expectation. Two deliberate mechanisms compose wrongly: the hop does not know the difference between "not yet tried" and "deliberately routed away from".

## Two ways to fix it, and I do not want to pick silently

**(a) Teach the hop about the rollover.** Record the account a rollover preempted off, and exclude it from the hop for that request. Narrow and faithful to both features — the hop still gets its one bounded move, and the rollover decision survives. Needs the server to learn something selection currently keeps to itself, which is what this branch's removed `decision` parameter existed for.

**(b) Accept the landing and update the test.** `a` genuinely has the most headroom at that moment, and refusing to use it while `b` is throttled costs availability to honour a heuristic. This is the cheaper change and the weaker one.

I lean (a): the rollover preemption is not advisory, and an availability hop that quietly reverses a routing decision made one step earlier is the kind of thing that is very hard to see in production.

Everything else on this branch is green.

ak2k:
Thanks for settling (a) versus (b) yourself, and for taking the hop fix. Scoping the preempted account to the synchronous walk rather than an instance field is the part that had to be right: the hop runs after a round trip, by which time another request has selected. Re-ran the flag-off differential against d4fbf12 — 0 divergences over 301,912 compared selections, and the flag-on control still fires.

Issue #286, opened 2026-09-06
The one-hop rate-limit failover moves the global current account, so the whole fleet follows a per-request hop


Frozen666:
#271's comment says the hop "is deliberately not a rotation policy: at most once per request". The hop itself is bounded as described — but the way it picks its destination rotates the fleet, and under a sustained 429 the current account bounces between two siblings with every hop.

## Observed (1.1.16, `expiryRouting` on, `distributeSessions` off)

```
19:48:41  Switched to account "B"
19:48:41  Rate-limit 429 on "A" — failing over once to idle account "B"
19:48:42  Second account rate-limited too — the limit looks IP-scoped, not per-account
19:48:42  Rate-limit 429 on "B" — waiting 60s, retrying same account (no switch)
```

Same second: the hop's destination lookup logs `Switched to account "B"` *before* the hop line. From then on every request from every session lands on B, because B is now `currentIndex`. A minute earlier the same sequence had moved the fleet A → B → A. An operator watching `currentAccount` sees it flip back and forth, and the switch line is the ordinary rotation message, so nothing says why.

(In this episode the "IP-scoped" reading turned out to be correct — idle accounts 429'd on their first request too — so the hop was not going to help either way. The point here is only the cursor.)

## Mechanism

`server.js`, the hop:

```js
const alt = accountManager.getActiveAccount(
  new Set([...ctx.tried, ...(ctx.rolledOff || []), account.index]),
  ctx.model, ctx.advisorModel, ctx.sessionId, ctx.provider,
);
```

`getActiveAccount` → `_pickActiveAccount` → `_select`. The current account is in `exclude`, so `_currentBarredOnlyFor` is false (it requires `!exclude?.has(current.index)`), `_select` falls to `_selectNext`, and `_selectNext` calls `_setCurrent(best)`. A lookup meant to answer "which idle sibling can take *this* request" has moved the fleet.

Two consequences beyond the visible flapping:

- **The hop works against its own premise.** It exists because "a sibling is sitting idle". After the hop, that sibling is current, so the burst that throttled A lands on B in full. Where the limit *is* per-account, that is what makes B 429 within a second — and the second 429 is then read as evidence of an IP-scoped limit, which forecloses a third try.
- **It interacts with what `_setCurrent` now does.** With `expiryRouting.preempt` on, `_setCurrent` takes a first-sight reading for rollover detection; a cursor that moves on every 429 hop takes readings the docs say a selection that merely *sends* a request must not take.

[...]

PR #287, opened 2026-09-06
feat(dashboard): warn when a session is getting nothing back

Frozen666:
Stacked on #289, which adds the signal this reads. The delta of this PR is the top commit only.

**Rewritten since the first version.** That one guessed from token totals and was wrong; an independent review measured three false positives and one false negative, and the whole approach was withdrawn. The history is in the comments below rather than force-pushed away, because the reasoning matters more than the diff.

## What it shows

A banner at the top of the page, empty unless something is wrong:

- **a session with `STARVED_MIN` (5) client requests in a row and nothing usable back** — named when `proxy.sessionDetail` is on, and reported unnamed via `sessions.starvedMax` when it is off, since that half must not be opt-in;
- **an account that needs a person** — a broken token, a disabled entry;
- **an account billing real money**, which no quota bar says.

Deliberately absent: a spent weekly bucket, a rate-limit back-off, and an upstream refusal. All three clear themselves, and a banner that is on every day is one nobody reads on the day it matters.

## What changed from the first attempt

- **It reads `starved` (#289) instead of guessing from `reports`.** The old heuristic accused three healthy shapes (`count_tokens`-only, repeated 4xx, a third-party upstream returning no `usage`) and missed a stream that dies after `message_start`. `starved` counts client-request outcomes at the one place that knows them.
- **No age floor.** The old one needed one because `reports === 0` is true of healthy sessions from birth. Five consecutive failures is true of no healthy session at any age, so a floor would only delay a true positive.
- **The `defaultTarget` wedged-fleet check is gone.** The review showed it fires on all-throttled and all-quota fleets that are still serving through the exhausted-fleet probe — the exact two states this PR argues should stay silent. A genuinely wedged fleet now surfaces as starving sessions instead, which is both true and actionable.
- **`entitlement` and `upstream-rejected` dropped** from the account list: both are self-clearing back-off, not something a person must act on. Only `error` and `disabled` remain.
- **The threshold is serialized into the page** alongside the helper that closes over it, with a test — otherwise the page would parse and then `ReferenceError` at first render.

## Testing

Tests build their payload from a real `AccountManager.getStatus()` and drive real sessions through `beginSession`/`endSession`. The previous version was validated against an object the server can never emit (`defaultTarget: null` with every `unavailable: null`), and that impossible fixture is precisely what hid the false positive.

Covered: a starving session named; one usable answer clearing it; below-threshold silence; a brand-new session; the unnamed path with `sessionDetail` off, including that it is not double-reported when both surfaces are available; each self-clearing account state staying silent while `error`/`disabled` report; real money reported but the mere ability to bill not; and the serialized threshold.

`npm run lint` clean. `npm test` **1347/1347 on Node 20.19, 22.19 and 24.12** — all three, after CI caught a teardown that only hung on 20 and 22.


Frozen666:
Marking this a draft and correcting it — an independent review found three things, one of which is my mistake and two of which make the detector wrong as written.

**The branch contained a change it never described.** Alongside the banner it carried a rewrite of how both failover hops pick their destination (a new `pickAlternate`, plus the 429 and 5xx call sites and a test) — work for #286 that a `git checkout -B` carried across from another branch and `git add -A` swept into a commit titled `feat(dashboard)`. Anyone approving "a warning banner" would have merged an unreviewed change to the 429/5xx path. Force-pushed; the branch is now `docs/usage.md`, `src/dashboard.js`, `test/dashboard.test.js` and nothing else. Apologies — that should never have reached you.

**The session detector fires on a healthy session.** `/v1/messages/count_tokens` answers `{"input_tokens": N}` with no `usage` object, so it records no report while still incrementing the session's request count — and Claude Code sends those constantly. Measured on this branch through a real proxy: six 200-OK `count_tokens` calls produce `requests = 6, tokens = {}` and a red "it is failing". A developer pasting files for two minutes before their first message gets accused. `reports === 0` cannot separate "answered, no usage object" from "never answered"; that needs a signal the tracker does not currently keep (any 2xx for the session), which is a server-side change rather than a dashboard one.

**The wedged-fleet check contradicts the rest.** `defaultTarget` comes from `previewRouteIndex`, which by its own JSDoc omits the exhausted-fleet probe fallback that `getActiveAccount` has. So on an all-throttled or all-quota fleet it is null while the proxy is still serving through the probe — a red "No account can serve an unrouted request right now" for exactly the two states this PR argues should stay silent. And my test for it passed only because the fixture was one `getStatus()` can never emit (`defaultTarget: null` with every `unavailable: null`).

Also fair, and I agree: `entitlement` and `upstream-rejected` are self-clearing back-off rather than "needs a person"; `requests` counts forward attempts (retries, reauths, hops), so the number in the banner — and the "117 requests" in the body above — is roughly double the developer's actual request count; and a stream that dies after `message_start` still records a report, so the likeliest shape of "two hours getting nothing" is the one case this misses.

I would rather not ship a banner that cries wolf on a common healthy pattern, so I am leaving this in draft. The idea stands and the motivating case is real — the session that prompted it made 117 attempts over two hours with no usage report and was invisible on the dashboard — but the honest version needs the tracker to record whether a session ever got a 2xx, and the fleet check needs to ask about the probe path rather than about `defaultTarget`. Happy to take it that way if you want it.

MagicalTux:
Verified the leaked failover rewrite is fully gone; the delta is the banner only. Pushed two small things to your branch on top of the #289 fix: `docs/usage.md` said the banner reports an account billing real money, while the code deliberately draws no spend line (and the test asserts it stays silent) — the doc now matches; and when every account is stalled for *different* reasons the banner named only the first one's, so it now says both. Merging after #289.

PR #289, opened 2026-09-07
feat(sessions): count consecutive client requests that got nothing back

Frozen666:
Groundwork for #287 (a dashboard warning), but useful on its own — `teamclaude status --json` gains the field either way. Independent of everything else open.

## The case

A session that is failing every request looks exactly like an idle one: both report zero tokens, both go quiet. On a six-account fleet one session ran **117 attempts over two hours with no successful answer**, and it was noticed only because the fleet started flapping around it. Nine healthy sessions on the same proxy were indistinguishable from it in the status payload.

## Why the token counters cannot answer this

I first tried `reports === 0` and it is wrong in **both** directions — measured against a real proxy, one row per shape:

| shape | client sees | `requests` | Σ`reports` | correct verdict |
|---|---|---|---|---|
| `count_tokens` only ×6 | 200 | 6 | **0** | healthy — `reports` says failing |
| repeated 400 ×6 | 400 | 6 | **0** | healthy — `reports` says failing |
| third-party 200, no `usage` ×6 | 200 | 6 | **0** | healthy — `reports` says failing |
| stream dies after `message_start` ×3 | broken | 3 | **3** | failing — `reports` says healthy |
| all accounts unavailable ×3 | 429 | **0** | 0 | failing — **invisible to any count** |

The last row is the one that matters most: `recordSession` sits below three early returns in `forwardRequest`, so a fleet that can serve nothing never increments the session's request count at all. It reads identically to a session that has never sent anything.

`count` is separately unusable as a denominator — it counts forward attempts, so retries, reauths and failover hops inflate it (3 client requests against a 5xx upstream measured as 6).

## The counter

`starved`: consecutive **client** requests that ended without a usable answer, reset to 0 by any usable answer.

A streak rather than a total, so a blip during an upstream wobble never accumulates on a session that is otherwise working, and no denominator is needed. A 4xx counts as an **answer** — it tells the client something true about what it sent, and a session getting legitimate 400s is working, not starving. A 429 is a refusal to answer and a 5xx is a failure to.

Recorded once per client request, in `endSession`, from a status classified at the three delivery paths. Abandonment is a third state: the proxy destroys the socket itself on a dead stream, so inferring "the client left" at the end would reclassify the worst failure as the user's own doing — it is observed where it happens instead.

`stats()` also reports `starvedMax` over **active** sessions, so a reader without `proxy.sessionDetail` still sees that something is starving, and so it clears itself two minutes after the session stops trying.

## Testing

Ten tests, each pinning one of the defects above rather than the happy path — the three false positives, the false negative (asserting `starved === 3` **and** `reports === 3` in the same test, since that contrast is the whole point), the attempts-vs-requests distinction, the no-account-available path, the abandonment third state, and the self-clearing aggregate. Most drive the real proxy against a stub upstream, because every defect in the first attempt lived at the call site rather than in the tracker.

`npm run lint` clean, `npm test` 1341/1341.


MagicalTux:
Read through every exit of `forwardRequest` — the outcome lands exactly once per client request and nothing in routing or retry changes. One fix pushed to your branch: `recordEarlyOutcome` went through `endSession`, whose `endRequest` releases an in-flight hold the early exits never took — with a sidecar request in flight that was the other request's hold, and its pin aged under it. Added `AccountManager.recordOutcome` (records the outcome, nothing else), used it from the early exits, and pinned it with a test. Also dropped a stale duplicate of the `finally` comment and moved the new helpers out from between `clientGone`'s JSDoc and `clientGone`. Merging once CI is green.

PR #290, opened 2026-09-07
feat(routing): adaptive distribution using plan tier and live congestion

nova28:
## Why

`distributeSessions` spreads new sessions by active-session count alone, which treats every account as interchangeable. That breaks down for mixed fleets:

- Accounts can be on different plans. An even split sends a small plan the same load as a much larger one.
- Even spreading fragments weekly windows. Concentrating work can finish a nearly spent window while preserving usable headroom elsewhere.

The existing modes offer either one active account or an even spread. Neither expresses “finish the account that is nearly done, but do not run it into the wall.”

## What

This adds a third value while preserving the existing `false` and `true` behavior:

```json
"distributeSessions": "adaptive"
```

Adaptive mode concentrates new sessions on the account with the least remaining weekly credit, tapers that preference near the configured switch threshold, and backs off when upstream congestion appears. Existing session pinning, priority ordering, and drain-on-disable behavior remain unchanged.

### Authoritative plan tiers, adaptive congestion

The router uses authoritative account/quota data where upstream provides it and learns only dynamic behavior:

| Input | Source/method | Purpose |
| --- | --- | --- |
| Weekly plan capacity | OAuth profile fields (`rate_limit_tier`, `seat_tier`, and organization type), normalized to the existing 1x/5x/20x weights | Compares remaining credit across differently sized plans without guessing the subscription |
| Tolerated concurrency | Back off after upstream 429s and grow cautiously after healthy responses | Prevents concentration from creating avoidable latency |

Quota response headers supply current utilization and reset time. Burn-rate observations are kept per quota window and only refreshed when that specific window appears in the response. Scoring uses utilization and burn reserve from the same binding window. If a future subscription tier is unknown, the whole competing tier falls back to utilization fractions instead of inventing a multiplier.

### Expiry routing composition

Adaptive selection uses `_bandedCandidates`, the same candidate set used by expiry routing. Expiry routing decides which windows are urgent; adaptive scoring decides which candidate to spend and how aggressively.

### Configured thresholds

The taper reaches zero at the operator's configured `switchThreshold`, including its per-bucket form.

## Auditable status output

In adaptive mode, `teamclaude status` reports the deterministic next target separately from the account's normalized score weight:

```
> account-a (Max 20x) (oauth, prio 0) active 3 sess (opus+ 2, sonnet 1)
  Weekly   [███████████░░░░░░░] 62% reset 3d8h
  Adaptive next · weight 36% of opus+  ·  3 sess / 3 inflight  ·  head 36.0% of 98%  ·  plan 20x  ·  conc 6.0
```

`next` is the account the deterministic picker would select for a new session. `weight` is the account's score normalized across the competing tier; it explains the decision but is not presented as a routing probability.

Session counts are also broken down by their pinned model-family bucket. The shared bucket is labeled `opus+` because Opus, Haiku, and unclassified models meter there together.

## Review fixes included

- A 429 no longer counts as a concurrency success, and the rejected admitted depth is retained for backoff.
- Burn-rate and concurrency learning remain attached to the correct credentials when an account is removed and indexes shift.
- Subscription capacity comes from persisted OAuth profile metadata rather than a volatile token-delta estimate.
- Burn-rate observations update only the quota windows refreshed by a response.
- Adaptive scoring uses a coherent binding quota window for utilization and reserve.
- Status diagnostics use the same expiry-routing candidate band as real selection.
- Status diagnostics exclude subscriptions belonging to another provider, matching request routing.
- String forms `off`, `false`, `no`, and `0` resolve to disabled distribution.
- Dynamic burn-rate and concurrency learning is persisted in the existing identity-keyed state file.
- Diagnostics distinguish deterministic selection (`next`) from normalized score explanation (`weight`).
- The real CLI adaptive round trip, renderer output, profile-tier weighting, fresh-window observation, 429 handling, provider partitioning, and account remapping have regression coverage.

## Testing

- Full test suite passes.
- Focused adaptive/status/CLI/quota/storm-control/failover tests pass.
- ESLint passes.
- `git diff --check` passes.

## Maintainer follow-ups (applied on this branch)

- `config.example.json` no longer opts into adaptive mode (`distributeSessions: false`).
- `adaptiveDistribution` is documented in `docs/configuration.md` and validated at startup; a bad value is a fatal, named error rather than a silent fall-through to even mode.
- `adaptiveStats()` reports per provider present (Codex accounts compete among themselves) and the status path memoizes it for 1 s.
- **Status output:** the `Serving …  cursor …` row and the session-following `>` marker apply only in adaptive mode; plain `distributeSessions: true` keeps the `Active` row exactly as before. Every account header gains a per-family session breakdown, e.g. `3 sess (opus+ 2, sonnet 1)`.
- The concurrency learner and the scorer now use one load definition (`sessions + inFlight`); `_pickAdaptive` applies the same band floor even mode does under `expiryRouting`; restored learner timestamps are clamped to now; an unrecognised `distributeSessions` string logs once.
- Note: the burn-rate and concurrency learners run and are persisted to the state file in every mode, not just adaptive (no routing effect unless adaptive is on).
- Rebased onto master after #311 (status strings sanitized via `nameText`/`safeLine`, numbers guarded).


src/account-manager.js
```diff
@@ -479,7 +512,12 @@ export class AccountManager {
   /** Release a slot taken by admit(). Safe to call once per successful admit. */
   release(index) {
     const account = this.accounts[index];
-    if (account && account.inFlight > 0) account.inFlight--;
+    if (!account) return;
+    // Counterpart to the throttle signal: a request that ran to completion at
+    // this depth is evidence the account sustained it. Read BEFORE the
+    // decrement, so it is the depth the request actually ran at.
+    this.concurrencyLearner.noteSuccess(index, account.inFlight || 0);
```
ak2k: An account throttled one request at a time never has its concurrency estimate lowered, so adaptive keeps sending it sessions. `release` runs in the `finally` around the upstream fetch, ahead of `pauseAccount`, so the success books the depth the request ran at and the throttle one lower. At depth 1 the throttle arrives at load 0 and is dropped. Twenty single-request 429s left the cap at its initial 6. Fix: capture the depth in `release` and pass it to `pauseAccount`; book a success only once the status is known.

src/account-manager.js
```diff
@@ -212,7 +233,19 @@ export class AccountManager {
...
+    // header and the remote dashboard all ask only that question.
+    this.distributionMode = distributionMode(distributeSessions);
+    this.distributeSessions = this.distributionMode !== 'off';
+    // Adaptive mode's two learned quantities: how big each account's weekly
+    // window is (its plan tier) and how much concurrency it tolerates. Both are
+    // inferred from live traffic — see adaptive-distribution.js. Constructed
+    // unconditionally and fed on every request regardless of mode, so switching
+    // to adaptive at runtime starts from what has already been observed instead
+    // of from a cold start.
+    this.capacityLearner = new CapacityLearner(adaptive);
```
ak2k: After a removal the survivors are routed against another account's plan tier and concurrency ceiling. Both learners key state by array position, and `removeAccount` renumbers while remapping the route pins, cursors, session pins and rollover observation, not these two. Removing index 0 of three left a 2,000,000 window reading 1,000,000, a cap of 6 reading 3.75, and a stale slot for the old index 2. Fix: shift both learners' keys beside `sessionTracker.remapAccounts`; `forget()` is defined on both learners and never called.

src/account-manager.js
```diff
@@ -2613,6 +2849,27 @@ export class AccountManager {
...
+   * selection, because what makes a sample usable is that it is FRESH: the
+   * learner divides tokens by the utilization move they caused, and only the
+   * moment a new reading arrives are the two known to cover the same interval.
+   */
+  _observeQuotaForLearning(account) {
+    if (!account) return;
+    const q = account.quota;
+    const now = Date.now();
+    this.capacityLearner.observeUtilization(account.index, 'unified7d', q.unified7d, now);
+    for (const { key } of FAMILY_WEEKLY_BUCKETS) {
```
ak2k: On mixed-family traffic the family window is never sized, so adaptive falls back to bare fractions and stops comparing plan tiers. This loop re-observes every family bucket the cached quota holds on every `updateQuota`, so a shared-only reading reaches the family bucket unchanged and `deltaU === 0` rebaselines it, zeroing pending tokens (adaptive-distribution.js:201). One interleaved shared-only reading took 1,000,000 pending tokens to 0 and left capacity null, against 20,000,000 for the control. Fix: observe only the buckets this update refreshed.

src/account-manager.js
```diff
@@ -2662,6 +2919,11 @@ export class AccountManager {
     // A request with no session id (or one the tracker has forgotten) is still a
     // real spend by the account, so the two scopes are recorded independently.
     this.sessionTracker.recordTokens(sessionId, bucket, usage);
+    // Third scope: what these tokens cost this account's weekly window. Unlike
+    // the two above, this one IS routed on — it is how adaptive distribution
+    // learns the account's plan tier (see adaptive-distribution.js). Fed in
+    // every mode so the tier is already known if adaptive is switched on later.
+    this.capacityLearner.recordTokens(accountIndex, bucket, usage);
```
ak2k: A family-heavy account is scored as though its shared weekly were half its real size, so sessions are pulled off it long before the window is finished. Tokens are booked to the request's own bucket only, while `_governingWeekly`'s comment says family spend meters into the shared weekly too. Against a true 100,000,000 shared weekly, half-and-half traffic learned 50,000,000 and family-only traffic left it null; the same-bucket control learned 100,000,000. Fix: credit family tokens to the shared bucket as well.

src/account-manager.js
```diff
@@ -900,9 +1032,104 @@ export class AccountManager {
...
+        maxRemaining,
+      });
+      r.score = score;
+      total += score;
+    }
+    for (const r of rows) {
+      // Share of the NEXT new session, not of the sessions already placed. A
+      // tier scoring zero everywhere (all inside their reserve) has no share to
+      // divide rather than an equal one.
+      r.share = total > 0 ? r.score / total : null;
```
ak2k: The percentages on the status line do not predict where sessions go, so an operator checking the rule against them is checking a number nothing implements. `share` is score over tier total, while `_pickAdaptive` returns the top-scoring account only (:929-944). An idle three-account fleet published 0.604/0.245/0.151, then took 17/3/0 of 20 new sessions. Fix: say in this comment, the renderer's and `docs/routing.md` that `share` ranks the accounts and the next session goes to the top one.

src/account-manager.js
```diff
@@ -900,9 +1032,104 @@ export class AccountManager {
...
+   * learned, but nothing is routing on them, and presenting them as if they
+   * governed anything would misreport what the proxy is doing.
+   */
+  adaptiveStats(model = null) {
+    if (this.distributionMode !== 'adaptive') return [];
+    const now = Date.now();
+    const bucket = this._weeklyBucketFor(model);
+    const threshold = this.thresholdFor(bucket);
+
+    const eligible = this.accounts.filter(a => this._isAvailable(a, model));
```
ak2k: On a mixed-provider fleet every published share is short by the other provider's slice, so an account taking all of its own provider's traffic reads 36%. The competing set is built from `_isAvailable` and priority only, with no provider term, while routing goes through `_excludeOtherProviders`. On a two-provider fleet both accounts read competing, shares 0.358 and 0.642 summing to 1, yet each provider's ten new sessions all went to its own account. Fix: apply the same exclusion here.

src/account-manager.js
```diff
@@ -8,10 +8,31 @@ import { SessionTracker } from './session-tracker.js';
...
+ *                               tapering at the threshold and backing off when
+ *                               the account is busy (see adaptive-distribution.js)
+ * An unrecognised string is treated as 'even' rather than rejected: it is
+ * plainly a request to distribute, and refusing to distribute at all would be
+ * the worse reading of a typo.
+ */
+export function distributionMode(setting) {
+  if (setting === 'adaptive') return 'adaptive';
+  if (!setting) return 'off';
+  return 'even';
```
ak2k, line 48: An operator who hand-writes `"distributeSessions": "off"` gets even distribution switched on, and the status line reports it as on. Every non-empty string that is not `adaptive` reaches this line: `off`, `false`, `no` and `0` all return `even`, and the reload path applies the raw disk value live (index.js:386-387), so it takes effect without a restart. The `distribute` CLI is safe: it stores the mapped value. These four are the off-words, not typos. Fix: map them to `off`.

ak2k (commented):
I read this because it overlaps work I had already been doing, and I wanted to see how it behaves before deciding what to do with mine.
I drove each point below against the branch at dd716e3 with a small script written from the claim alone, so every number in the comments comes from a run.
The new tests exercise each piece on its own; the defects below are in how the pieces fit together. The concurrency test sets `inFlight` by hand and calls `pauseAccount`, so nothing in the suite runs `admit`, `release` and `pauseAccount` in the order the server runs them, which is the first comment below.
These are offered to help the feature land well.
Beyond these seven, I checked how the mechanisms behave, not whether the proxy should learn plan size from headers at all; that is a call for the maintainer. As it stands, on mixed traffic neither of the two things the title says it learns comes out right (comments 1, 3 and 4). Each is a contained fix inside the shape as written. The learner keys and the missing persistence go against choices the codebase already made. `removeAccount` remaps every other index-keyed structure, and the learners can follow it. The quota the estimates read is saved every minute through `exportQuotaState`, while the estimates live only in memory, so a proxy restarted daily never learns.

nova28:
@ak2k Thank you for the thorough, evidence-backed review. We have addressed the seven integration issues you identified. We also decided not to infer subscription capacity from token deltas: adaptive routing now uses the persisted OAuth profile tier (1x/5x/20x), uses quota headers for utilization and reset data, and learns only dynamic burn rate and concurrency. That learned state is persisted by account identity across service restarts, while an actual quota rollover rebaselines the burn observation. Unknown tiers fall back to utilization-based scoring rather than a guessed multiplier. The PR and regression coverage have been updated accordingly.

MagicalTux (dismissed):
Read in full, including the earlier seven-point review — all seven are addressed in code, and the `false`/`true` paths are unchanged (`_pickLeastLoadedEven` is byte-identical). The mode itself is fine as an opt-in. Four things before merge:

1. **`config.example.json` sets `"distributeSessions": "adaptive"`.** The help text points new users at that file, so anyone copying it opts into the new mode. New behaviour is off by default in this project; drop the line or set it to `false`.

2. **`adaptiveDistribution` is an undocumented, unvalidated config key** (index.js spreads it raw over `ADAPTIVE_DEFAULTS`). Either remove the passthrough, or document it in `docs/configuration.md` and validate the fields (e.g. `burnWindowMs: 0` yields an Infinity burn rate today).

3. **`adaptiveStats()` hard-codes `DEFAULT_PROVIDER`**, so on a fleet with Codex accounts every Codex subscription renders "(not competing)" with no `next`, although Codex requests are routed adaptively among them. Iterate the providers present, or state the limitation in `docs/routing.md`.

4. **Status text changes for users who did not opt in.** With plain `true` the "Active" row becomes `Serving …  cursor …` and the `>` marker moves; with any mode the header gains `(opus+ N, sonnet N)`. Both may be improvements, but neither is in the PR body, and the "Active" change breaks anyone grepping it. Describe both in the body, or gate "Serving" to adaptive mode.

One more sentence for the body: the burn-rate and concurrency learners run and are persisted to the state file in every mode, not just adaptive (no routing effect, but users who never opt in now carry learned state on disk).

PR #291, opened 2026-09-07
fix(probe): never send a third-party backend's key to Anthropic

aramalipoor:
`Prober.probeAll` selected accounts with `account.type === 'oauth'`, and `fetchUsage` posts to a fixed `https://api.anthropic.com/api/oauth/usage`.

A third-party backend account is typed `oauth` with a static token — that is the documented pattern in [accounts.md](docs/accounts.md#third-party-backend-accounts), which is how DeepSeek and GLM are meant to be added. So with the probe enabled, every cycle sent **that provider's API key to api.anthropic.com**.

Observed on a fleet with a DeepSeek backend and one other Anthropic-compatible provider, both added exactly as documented:

```
deepseek (oauth, prio 200) active
  Probe    error 2m ago, 55ms, HTTP 429: Rate limited. Please try again later.
```

Reproduced directly: a third-party key against `/api/oauth/usage` returns `HTTP 429`.

Two costs:

- **The key leaves the machine for a party it was never issued to**, on a timer, for as long as the probe is on.
- The 429 is recorded as a permanent probe error against an account that is serving traffic perfectly well — so the one row an operator reads to judge a backend's health says it is broken when it is not. Routing is unaffected; only the readout and the outbound request are wrong.

## The fix

Probing now requires the account to carry no `upstream`. That is the same line `warmer.js` already draws in `_isWarmTarget`: an account with its own base URL has no Anthropic subscription usage to read. The rule moves into a named `_isProbeTarget` so the reason is stated once.

`index.js` built its pre-probe status view with the looser test (`type === 'oauth'`) while the warm view beside it already excluded `upstream`. It now matches, so a backend reads `not-applicable` instead of pending a probe that will never come.

One regression test: a fleet of one Claude account and one `upstream` account, asserting only the Claude token is ever handed to the probe function, and that the backend's quota is left untouched.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,6 +64,7 @@ Volatile runtime state (observed quota) is written separately to `teamclaude.sta
 | `accounts[].accountUuid` | Anthropic account (person) id; set automatically from the OAuth profile |
 | `accounts[].orgUuid` / `orgName` | Organization the account is scoped to — lets one email hold multiple org accounts |
 | `accounts[].priority` | Rotation preference, lower = preferred (default 0) |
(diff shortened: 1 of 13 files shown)


aramalipoor:
Pushed a follow-up: `getStatus` used the same loose `type === 'oauth'` test, so a skipped account read `never` — indistinguishable from a first cycle that has not run. It now reads `not-applicable`, and all three call sites share `_isProbeTarget`.

Note this branch alone shows one red test, `a probe that reports the family still spent keeps it spent`. It is unrelated (a fixture pinned to a calendar date that has passed) and is fixed in #292; with both applied the suite is green at 880.

PR #292, opened 2026-09-07
fix(test): date resets relative to now, not by calendar

aramalipoor:
`master` is red today, and has been since 2026-09-04 passed.

The "a probe that reports the family still spent keeps it spent" case (mine, from #240) pinned `resets_at` to the absolute date `2026-09-04T10:00:00Z`. Once that moment went by, `_clearExpiredQuotas` did exactly the right thing — dropped a reading whose window had closed — and the account read available for Fable. The assertion that fails is the availability one, so the failure presents as a Fable routing regression rather than as a stale fixture:

```
✖ a probe that reports the family still spent keeps it spent
  + { ...account object... }   - null
```

I traced it to a live gate before spotting the fixture, which is the cost this kind of test imposes on the next person.

The window is now `Date.now() + 24h`, so the test asserts the behaviour it was written for on any day it runs. The other absolute dates in the file are parse-and-map assertions with no expiry logic behind them, and are left alone.

Unrelated, noticed while verifying: `test/activity-entries.test.js:71` ("a request aborted mid-body closes its activity entry") fails intermittently under full-suite load with "no activity entry was opened, so this proves nothing", and passes 3/3 in isolation. Not touched here.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -64,6 +64,7 @@ Volatile runtime state (observed quota) is written separately to `teamclaude.sta
 | `accounts[].accountUuid` | Anthropic account (person) id; set automatically from the OAuth profile |
 | `accounts[].orgUuid` / `orgName` | Organization the account is scoped to — lets one email hold multiple org accounts |
 | `accounts[].priority` | Rotation preference, lower = preferred (default 0) |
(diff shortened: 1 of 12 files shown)


aramalipoor:
Second instance pushed: `scoped-weekly.test.js` (#231) asserts `scoped.opus.resetAt > Date.now()` against a fixture pinned to `2026-09-05`, so it fails the same way on its own schedule. Both fixtures are now relative. Full suite green apart from the intermittent `activity-entries.test.js:71` noted above.

aramalipoor:
Superseded — upstream fixed both fixtures independently: `quota-probe.test.js` now derives `at` from `Date.now()`, and `scoped-weekly.test.js` asserts the exact parsed value instead of futurity. Nothing left to land here. #291 still applies.

PR #293, opened 2026-09-07
feat(quota): show a third-party backend's own quota

aramalipoor:
A third-party backend account has no Anthropic quota, so every bar reads `unknown`:

```
  deepseek (oauth, prio 200) active
  Quota    unknown
```

That hides the one number deciding whether the fallback will actually answer when the Claude accounts are done.

## Where the provider-specific part lives

Providers publish that number in their own shapes, at their own paths, and most publish nothing at all. So all of it is confined to one new module, `src/backend-quota.js` — a host-matched entry with a path and a `parse`, returning a normalized reading:

```js
{ label: 'Balance', text: '$25.81', utilization: null }
```

Everything else stays generic and names no vendor:

| Layer | Knows |
|---|---|
| `backend-quota.js` | every provider detail — **the only such file** |
| `prober.js` | "some accounts have a backend reading"; schedules it on the existing interval, records ok/error in the same status row |
| `account-manager.js` | one field, `quota.backend`, beside — never mixed into — the unified buckets, which model Anthropic's utilization alone |
| `status-renderer.js` | draws a bar when `utilization` is set, the text when it is not; no idea which provider it is drawing |

Adding a provider is one entry in `PROVIDERS`. Nothing else changes.

## What it refuses to do

- A provider with no entry keeps reading `unknown`. Nothing is inferred from a shape we have not seen.
- A failed read leaves the last value in place and records the error — a 500 is not evidence of a new balance.
- The reading is not persisted: money moves, and a figure restored from disk would be presented as current.
- It is not folded into the switch threshold. It is a fact for the operator, not a gate; making it one would need a policy this PR has no data to choose.

## First entry

DeepSeek: `GET {upstream-origin}/user/balance` → `Balance $25.81`. The path resolves against the account's configured `upstream`, so a regional or proxied host keeps working. Verified against a live account.

`test/backend-quota.test.js`, 10 tests: host matching (including subdomains and malformed URLs), normalization, the request's URL and credential, the `is_available: false` case, each failure mode, the prober routing a backend away from the Anthropic endpoint, a failed read preserving the last value, status-row applicability, and the renderer drawing both shapes.

One assertion in `quota-probe.test.js` (from #291, still open) moves to an upstream nothing publishes for — a DeepSeek backend is now probeable, so `not-applicable` is no longer the right answer for it.

`npm test` 1343 pass, `npm run lint` clean.


aramalipoor:
Note on scope: this branch also carries #291 (the probe no longer sends a third-party backend's key to `api.anthropic.com`, which answers 429), since #291 was closed. That is why `src/index.js` and part of `src/prober.js` appear here. The two are separable if you would rather take them apart — say which and I will split it.

MagicalTux:
Reviewed in full — design is right (probe-gated, key only to the configured origin, nothing persisted). I rebased this onto master now that #294 is in, resolved the `prober.js` / `quota-probe.test.js` overlap, and pushed three small fixes to your branch: `fetchBackendQuota` now defaults to `proxyFetch` (the OAuth usage probe already does, so a host behind a configured upstream proxy gets a reading instead of a permanent error); the "never sends a third-party key" test stubs `backendFn` (the DeepSeek account is a backend target now, and the default read left the suite for the network); and `_isBackendTarget` moved below `_isProbeTarget` so the JSDoc documents the method it was written for. Merging once CI is green. One nit for later, not blocking: `providerFor` matches on `host`, so an upstream with a non-default port silently gets no reading — `hostname` would fix it.

PR #297, opened 2026-09-07
fix(third-party): strip undocumented cache_control subfields for custom upstreams

aramalipoor:
- Claude Code sends a `scope` subfield inside `cache_control` that Anthropic accepts but strict Anthropic-compatible backends reject with `400 unknown parameter system.cache_control.scope`, breaking every request once such an account is selected
- Custom-upstream legs now forward only the documented `type`/`ttl` (system, message, and tool blocks); first-party legs pass through byte-identical
- Dropping a cache hint can only cost a cache hit, never correctness; `stripRequestFields` stays top-level-only, no config needed for this
- Verify: new unit + wiring tests green (14), eslint clean, full suite shows no new failures vs base


MagicalTux (dismissed):
Thanks — the wiring is right (the strip only runs under `rewritesBody(account)` and `account.upstream`, Content-Length is refreshed, sx-routed first-party accounts are untouched). Three things before this can go in:

1. **`ttl` is stripped, contradicting the description.** `KEPT_SUBFIELDS` in `src/cache-control-sanitize.js` keeps only `type`, but the PR body says `type`/`ttl` are forwarded, and `ttl` is documented by Anthropic (`"5m"` | `"1h"`). On a mirror that honours `ttl: "1h"` this silently drops cache lifetime to 5 minutes. Keep `ttl`, and drop the docs line that calls it unsupported (that generalises one backend's behaviour to all of them).

2. **The walk rewrites any object with a key named `cache_control`, not just cache hints.** `stripUnknownSubfields` descends into every object, so a `tool_use.input` (or `tool_result` content, or `metadata`) that happens to carry a `cache_control` field gets edited or emptied — that changes a tool call the model already emitted. Please walk only the documented positions: root, `system[]`, `messages[].content[]` (and `tool_result.content[]`), `tools[]`.

3. **Any `upstream` account is affected, including first-party relays.** `docs/accounts.md` defines `upstream` as the base URL of the target API; an Anthropic OAuth account pointed at a regional mirror or egress relay would now lose `scope` on every request. Every other body rewrite keyed on `upstream` (`modelMap`, `stripRequestFields`) is opt-in. Either make this opt-in too (e.g. let `stripRequestFields` take a nested path like `"cache_control.scope"`), or state plainly in the body and in `docs/configuration.md` that every `upstream` account gets it.

A test that `tool_use.input` is left byte-identical would pin (2).

PR #301, opened 2026-09-07
Fix proxy scheduling and harden request lifecycle under overload

nova28:
## Summary

Fix macOS scheduling starvation and harden proxy request lifecycles under concurrent load. Adaptive-routing PR #290 remains separate.

- Use Interactive LaunchAgent scheduling instead of Background. A controlled same-work comparison on the affected, resource-contended host took 5,141 ms versus 90 ms; this identifies a scheduling contributor, not a general throughput gain.
- Add a bounded CLI status deadline, event-loop diagnostics, and ingestion/upstream occupancy in status JSON.
- Bound uploads and buffered responses; cancel queued work, upstream reads and quota-hold timers on disconnect. Close activity entries exactly once on all tested paths.
- Add explicit bounded upstream admission, separate queue/headers deadlines, and 503 with Retry-After on local overload without account rotation. Preserve the original 256-socket default so long streams do not serialize ordinary concurrent sessions.
- Keep ingestion concurrency opt-in. Add HTTP/2 stream-isolation and overload/recovery tests, safe structural metadata rewriting, shared agent guidance and a reliability runbook.

## Validation

- Standalone PR suite: **1,363 passed**, zero failed/skipped/cancelled on Node 20.19.4, 22.22.0 and 24.13.0. Fresh post-review full-suite run also passed all 1,363 tests.
- Targeted hardening suite: 40 passed. ESLint and git whitespace checks passed.
- TDD covered queue expiry/cancellation, rejected/disconnected activity cleanup, slow/oversized uploads, silent and oversized buffered responses, upstream disconnects, quota-hold cancellation, and metadata ambiguity. A burst regression caught h1 teardown hiding the rejection response.
- Twenty repeated recovery runs: 1,200 one-MiB requests, 500 successful status probes, HTTP/2 isolation and queue drain checks.
- Continuous 60-second isolated soak: **1,608 requests accounted for**, **585/585 status probes passed**, p95 **5.92 ms**, zero event-loop stalls. After one synthetic account began returning 429s, both unaffected pinned accounts continued progressing (536 successful requests each). The throttled account had 328 successes, four 429s and 204 expected client cancellations during its existing rate-limit pause; all activity entries closed.
- Soak memory samples after explicit GC: heap approximately 11–12 MiB; sampled RSS up to 311 MiB for the combined proxy, fake upstream and load generator. RSS levelled off near the end; this is not a long-term leak guarantee or universal memory bound.

## Local deployment

The hardening is deployed locally alongside adaptive routing as commit `bfc6e9a`; that combined checkout passed **1,418 tests** before restart. Post-restart checks: 30/30 status probes succeeded, median 2.3 ms, and zero event-loop stalls. A later check after over five minutes still showed zero stalls, Interactive scheduling, and no queued upstream requests. The ingress gate remains disabled by default. This PR's final cleanup changes comments and documentation only; no additional service restart was needed.

## Scope and operational follow-ups

Privacy review: checked added content across the final diff and all PR commits, commit messages, and PR text for personal identifiers, private paths and credentials. No such data was found in the reviewed additions; fixtures use synthetic values and operational evidence is aggregated. Shared agent guidance now requires this check before publishing or marking future PRs ready. Normal Git contributor attribution is unchanged.

This is ready for review/merge as scoped; it is not a guarantee against upstream outages. Keep ingestion admission opt-in and observe multi-hour real workloads before broader tuning. Its limits do not cap total sessions, retained retry bodies or whole-process memory. OAuth/Remote Control/attachments/forward-proxy/WebSocket relays have separate paths; the legacy global-fetch escape hatch bypasses Node-transport admission.

Upstream 429 pauses can still delay first bytes while local status is healthy. The synthetic pinned-account test does not establish adaptive-routing fairness. Optimal yield batch size, sustained production memory behavior and alert thresholds remain operational tuning work, not claims made by this PR. See `docs/reliability.md` for limits, incident checks and rollout guidance.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -72,3 +72,15 @@ and do not restart the live service just to experiment without authorization.
 - Do not immediately destroy an h1 request with unread upload bytes after
   writing a rejection: the reset can hide the 413/503 from its caller. Teardown
   must be bounded, while h2 rejection must close only the affected stream.
+
+### Privacy before publishing or merging
+
+- Audit the final diff, every new commit's content, and the PR description for
+  personal data and secrets before publishing or marking a PR ready. Removing
+  data only from the final tree does not remove it from earlier commits.
+- Keep real account emails/IDs, credentials, session identifiers, private local
+  paths, request payloads and raw operational logs out of source, fixtures and
+  PR text. Use clearly synthetic accounts and sanitized aggregate measurements.
+- If sensitive data was already published, stop and report it without repeating
+  the value. Coordinate credential revocation and history cleanup as appropriate;
+  do not silently force-push or rewrite contributor attribution.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,74 @@
+# Agent guidance
+
+## Proxy reliability: preserve these findings
+
+TeamClaude is operational infrastructure. Diagnose with live evidence before
+changing scheduling, routing, timeouts, or concurrency. Preserve unrelated work
+and do not restart the live service just to experiment without authorization.
+
+### macOS service scheduling
+
+- Keep the user-facing HTTP LaunchAgent's `ProcessType` set to `Interactive`
+  in `src/service.js`. Do not change it to `Background` as a daemon convention.
+  Background scheduling was demonstrated to starve this proxy under host
+  resource pressure, causing local status timeouts and large event-loop lag.
+- `Adaptive` launchd scheduling relies on XPC activity for promotion; ordinary
+  HTTP clients do not provide that signal. This is unrelated to TeamClaude's
+  adaptive account-routing mode.
+- Preserve the service-rendering regression assertion in `test/service.test.js`.
+  A source change alone does not update an installed LaunchAgent: deployment
+  must regenerate its plist and reload the service, then verify the effective
+  policy with `launchctl print gui/$(id -u)/com.karpeleslab.teamclaude`.
+
+### Probe the correct layer
+
+- Use `teamclaude status` or `GET /teamclaude/status` for local proxy status.
+  **Do not use `/status`: it is forwarded upstream, not a local health check.**
+  Respect configured authentication; never expose credentials in diagnostics.
+- Separate connection refusal, local status timeout/event-loop lag, and upstream
+  first-byte latency. They are different failure modes, not interchangeable
+  evidence of an Anthropic outage.
+- Upstream 429s and quota retry holds can delay API responses while local status
+  remains healthy. An account's `active` credential state does not necessarily
+  mean it is eligible for routing below the configured quota threshold.
+- Before attributing a stall to JavaScript hot loops, inspect process CPU time
+  alongside wall time, host memory pressure, effective service scheduling,
+  event-loop metrics, and time-correlated logs. A sampled busy stack alone does
+  not prove that the process is receiving enough CPU time.
+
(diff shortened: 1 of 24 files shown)


MagicalTux (changes requested):
Thanks for the depth of work here. The core — bounded, cancellable upstream admission; cancelling quota-hold timers and upstream reads on disconnect; leaving the Background QoS clamp — is sound and the activity-entry accounting reads right. But it can't go in as one PR. Please split it (roughly: service.js ProcessType; upstream admission + disconnect cancellation; ingress bounds; status/event-loop diagnostics), and drop the parts below.

**Drop from the PR**
- `CLAUDE.md`, `AGENTS.md`, `docs/reliability.md`. This repo deliberately carries no agent-guidance files, and the reliability doc is one host's incident narrative. The env-var table belongs in `docs/configuration.md`; the rest doesn't belong in-tree.
- The `src/model.js` and `src/account-uuid-rewrite.js` rewrites. They aren't described in the body, fix no behavioural bug (the "nested metadata" test already passes on master), and both regress: `parseRequestModel` now `JSON.parse`s the whole body on every request when no model hook is installed — which is exactly the headless LaunchAgent this PR is about — where master's finder stops after the first key. The uuid patcher goes from a byte-exact streaming patch to parse + regex re-tokenise + re-serialise on every forward attempt including retries.

**Defects in what stays**
- `src/server.js` catch block: the new `if (clientGone(res)) { ctx.status = 499; return; }` returns before the request-log stream is ended, so with `--log-dir` every mid-stream disconnect leaks one fd.
- `src/upstream-fetch.js`: `AbortSignal.any` needs Node ≥ 20.3; `package.json` says `>=20.0.0`.
- `res.once('close')` → abort → `ctx.status = 499` in `finally` relies on the finally running before the nextTick'd 'close' after a normal `res.end()`. Guard on `!res.writableEnded` so a future await after `end()` doesn't flip successes to 499.
- `writeRequestBody` yields a `setTimeout(1)` after every 64 KiB chunk even when the socket has room: ~60 ms+ per 4 MB context, per attempt. Yield only on backpressure.
- `relayFair` on the Remote Control / attachment relay path is undescribed and adds a stall per 8 chunks to long-polls and image downloads.

**Defaults that change behaviour for everyone** (need to be off by default or described prominently): 32 MiB body cap → 413 and a 120 s absolute upload deadline → 408 (also on `/v1/files`, which passed through before); 32 MiB buffered-response cap (affects `/v1/files/{id}/content`); per-origin queue of 64 with 5 s wait → 503. Anyone who lowered `TEAMCLAUDE_UPSTREAM_MAX_SOCKETS` per the old docs now gets 503s instead of queueing, and the 503 is attributed to the account in activity output.

On `ProcessType`: leaving `Background` is justified. `Standard` is the conservative pick for a daemon; `Interactive` grants app-level priority. Either way, say in the body that existing installs are unaffected until `service install` is re-run.

MagicalTux:
Landed as three PRs per the review here: #320 (ProcessType → Standard), #321 (bounded upstream admission + disconnect cancellation, with the listed defects fixed), #322 (event-loop diagnostics + status deadline). The model.js / account-uuid-rewrite.js rewrites and the AGENTS.md / CLAUDE.md / docs/reliability.md additions were dropped as discussed. Thank you for the work — authorship is preserved on the ported commits.

PR #317, opened 2026-09-08
fix(server): gate the WebSocket upgrade like every other entry point

### LINKED ISSUE #314: The WebSocket upgrade handler is registered outside the API-key gate
`server.on('upgrade', …)` is registered outside the API-key gate, so a WebSocket handshake reaches `relayUpgrade` without presenting `proxy.apiKey` or a `clientKeys` entry. On a deployment bound to a public hostname that is an open relay to the configured upstream.

Measured against a live 1.1.17 behind a reverse proxy that forwards `Upgrade` — same URL, same absence of a key:

```
plain GET                  -> HTTP 401
genuine upgrade handshake  -> socket held open, relayed, no auth
```

## Why it is not a credential leak, and still worth closing

`relayUpgrade` forwards the client's own headers and injects nothing:

```js
for (const [key, value] of Object.entries(req.headers)) {
  const lk = key.toLowerCase();
  if (lk.startsWith(':') || lk === 'host' || lk === 'x-api-key') continue;
  headers[key] = value;
}
```

Note the `x-api-key` strip: the client's credential to *us* is deliberately not passed on. No `applyAuthHeaders`, no account selection, no pooled token — the caller authenticates as itself. So the exposure is not the fleet's credentials; it is that an unauthenticated third party gets a relay to `api.anthropic.com` (or whatever `upstream` names) carried on the operator's host, address and bandwidth. With `sx` provisioned and `useByDefault()` on, it also rides the operator's metered residential egress.

Every other path is gated: `requestHandler` runs `resolveClientAuth` before anything, `CONNECT` has the 407 gate, and the control plane adds the same-origin check on top. The upgrade handler is the one entry point that skips all of it. `GET /teamclaude/dashboard` is also pre-gate but is deliberately a static asset with no data, which is a different thing.

This is long-standing rather than new — identical in 1.1.16 and earlier; noticed while reviewing the 1.1.17 delta.

## Suggested fix

```js
server.on('upgrade', (req, socket, head) => {
  const auth = resolveClientAuth(config.proxy, req.headers['x-api-key']);
  if (!auth.ok && !isLoopbackAddr(req.socket.remoteAddress)) {
    socket.write('HTTP/1.1 401 Unauthorized\r\nConnection: close\r\n\r\n');
    socket.destroy();
    return;
  }
  relayUpgrade(req, socket, head, upstream, sx);
});
```

If the intended consumer (Remote Control / `teamclaude attach`) cannot set `x-api-key`, the key could ride `Sec-WebSocket-Protocol` or a query parameter — but it should ride something.

Worth noting for operators: a reverse proxy that sets `websocket: true` purely to get `proxy_http_version 1.1` and long read timeouts for streaming responses also forwards `Upgrade`, and so exposes this without anyone intending to.

Frozen666:
Closes #314.

`server.on('upgrade', …)` is registered outside the API-key gate, so a WebSocket handshake reaches `relayUpgrade` without presenting `proxy.apiKey` or a `clientKeys` entry. Measured against a live 1.1.17 behind a reverse proxy that forwards `Upgrade` — same URL, same absence of a key:

```
plain GET                  -> HTTP 401
genuine upgrade handshake  -> socket held open, relayed, no auth
```

(An HTTP/2 client gives a misleading 401 here, because `Upgrade` is meaningless in h2. It needs `--http1.1` to reproduce.)

## Severity, stated precisely

Not a credential leak. `relayUpgrade` forwards the client's own headers and injects nothing:

```js
for (const [key, value] of Object.entries(req.headers)) {
  const lk = key.toLowerCase();
  if (lk.startsWith(':') || lk === 'host' || lk === 'x-api-key') continue;
  headers[key] = value;
}
```

Note the `x-api-key` strip: the client's credential to *us* is deliberately not passed on. No `applyAuthHeaders`, no account selection, no pooled token — the caller authenticates as itself, so this is not a way to spend the fleet's quota. It is a way to reach the upstream **on the operator's host, address and bandwidth**, and with `sx` provisioned and `useByDefault()` on, on their metered residential egress. Every other entry point is gated: `requestHandler` runs `resolveClientAuth` first, `CONNECT` has the 407 gate, and the control plane adds the same-origin check. `GET /teamclaude/dashboard` is also pre-gate but is deliberately a static asset with no data — a different thing.

## The fix

`resolveUpgradeAuth` mirrors `resolveConnectAuth`: same `resolveClientAuth` comparison, same loopback exemption — so a local `teamclaude attach` is unaffected, and a deployment with no keys configured stays open by choice.

`x-api-key` only. A browser cannot set that header on a WebSocket handshake, so a browser client cannot authenticate here — deliberately. Reading the key out of `Sec-WebSocket-Protocol` instead would be worse than not supporting browsers: `relayUpgrade` **forwards** that header (it strips `x-api-key`, which is exactly why the handshake carries no operator credential today), so the key would reach the upstream; the offer list is attacker-sized, turning one guess per connection into thousands where `resolveConnectAuth` tries at most two; and the proxy cannot honour the negotiation anyway, because it relays the handshake rather than answering it. A test pins the refusal so the idea cannot quietly return. The first version of this PR did accept the subprotocol and did leak the key — see the comment below.

The refusal is written to the socket rather than dropped — a silently held socket is the shape of an outage, and that is exactly how this went unnoticed.

The MITM listener wires the same relay onto its own terminating server, which sits behind the CONNECT 407 gate. Already authenticated, so it is unchanged.

## Testing

The gate's truth table against a stubbed non-loopback address (no key, shared key, client key, wrong key, loopback exempt, no-keys-configured), the subprotocol path, and — because a unit test cannot show the handler is *wired* — a real socket handshake against a real proxy with `remoteAddress` stubbed to force the non-loopback branch: refused with `401` and never reaching the upstream without a key, relayed with one. Mutation-checked: disabling the gate turns all three red.

`npm run lint` clean. `npm test` **1508/1508 on Node 20.19, 22.19 and 24.12.**

## Note for operators

A reverse proxy that sets a websocket flag purely to get `proxy_http_version 1.1` and long read timeouts for streaming responses also forwards `Upgrade`, and so exposes this without anyone intending to. That is how ours was configured, and there was no way to keep the timeouts without the forwarding — which is what made fixing it here rather than there the only option.

### DIFF
--- a/src/server.js
+++ b/src/server.js
@@ -1230,10 +1230,24 @@ function relayStream(req, res, upstream, sx) {
  */
 export function resolveUpgradeAuth(req, socket, proxyConfig) {
   const auth = resolveClientAuth(proxyConfig, req?.headers?.['x-api-key']);
+  if (auth.ok) return auth;
   // Loopback is exempt from the key requirement, exactly as the HTTP and
-  // CONNECT gates are.
-  if (!auth.ok && isLoopbackAddr(socket?.remoteAddress)) return { ok: true, client: null };
-  return auth;
+  // CONNECT gates are — with the request path's two conditions on top, for
+  // the same actor: a web page in the operator's browser. A page can open a
+  // WebSocket to 127.0.0.1 with no CORS check at all, and its handshake is
+  // loopback-sourced too. What it cannot forge is `Origin`, which a browser
+  // sets on every handshake and a CLI never sends, nor `Host`, which a
+  // rebound name (attacker.example → 127.0.0.1) leaves naming the attacker.
+  if (!isLoopbackAddr(socket?.remoteAddress)) return auth;
+  const bindHost = proxyConfig?.host;
+  const origin = req?.headers?.origin;
+  if (origin) {
+    let originHost;
+    try { originHost = new URL(origin).host; } catch { return auth; }
+    if (!isLocalHostHeader(originHost, bindHost)) return auth;
+  }
+  if (!isLocalHostHeader(req?.headers?.host, bindHost)) return auth;
+  return { ok: true, client: null };
 }
 
 /**
--- a/src/server.js
+++ b/src/server.js
@@ -457,7 +457,26 @@ export function createProxyServer(accountManager, config, hooks = {}, sx = null,
   // call — Node fires 'upgrade' for that handshake, never 'request', so it
   // needs its own listener (base-URL routing path; the MITM path wires the
   // same relayUpgrade onto its own terminating server in mitm.js).
-  server.on('upgrade', (req, socket, head) => relayUpgrade(req, socket, head, upstream, sx));
+  server.on('upgrade', (req, socket, head) => {
+    // The upgrade handshake never reaches requestHandler, so it does not
+    // inherit the key gate above — it has to ask for itself. Without this a
+    // WebSocket handshake is an unauthenticated relay to `upstream`: the
+    // handshake carries no pooled credential (relayUpgrade forwards the
+    // client's own headers), so it is not a way to spend the fleet's quota,
+    // but it is a way to reach the upstream on this host's address and
+    // bandwidth. A deployment on a public hostname hands that to anyone.
+    if (!resolveUpgradeAuth(req, socket, config.proxy).ok) {
+      // Logged as well as answered: a WebSocket client discards the status
+      // line, so the 401 alone leaves an operator with a channel that is
+      // silently dead — the same shape as the outage this gate could cause if
+      // a client turns out not to send the key.
+      console.log(`[TeamClaude] WebSocket upgrade refused (no proxy key) from ${safeLine(socket?.remoteAddress || 'unknown')} for ${safeLine(req.url)}`);
+      try { socket.write('HTTP/1.1 401 Unauthorized\r\nConnection: close\r\n\r\n'); } catch { /* already gone */ }
+      socket.destroy();
+      return;
+    }
+    relayUpgrade(req, socket, head, upstream, sx);
+  });
 
   return server;
 }
@@ -1189,6 +1208,32 @@ function relayStream(req, res, upstream, sx) {
 
   if (['GET', 'HEAD'].includes(req.method)) upstreamReq.end();
   else req.pipe(upstreamReq);
+}
+
+/**
+ * The key gate for a WebSocket upgrade, in the shape of the CONNECT one.
+ *
+ * Separate from `resolveClientAuth` only because the answer depends on the
+ * socket's address as well as the header, and separate from the request path
+ * because `server.on('upgrade')` is a different event that no part of
+ * `requestHandler` runs for.
+ *
+ * `x-api-key` only. A browser cannot set that header on a WebSocket
+ * handshake, so a browser client cannot authenticate here — deliberately.
+ * The obvious alternative, reading the key out of `Sec-WebSocket-Protocol`,
+ * is worse than not supporting browsers: relayUpgrade forwards that header to
+ * the upstream (it strips `x-api-key`, which is the whole reason the
+ * handshake carries no operator credential today), the offer list is
+ * attacker-sized so it turns one guess per connection into thousands, and the
+ * proxy cannot honour the negotiation anyway because it relays the handshake
+ * rather than answering it.
+ */
+export function resolveUpgradeAuth(req, socket, proxyConfig) {
+  const auth = resolveClientAuth(proxyConfig, req?.headers?.['x-api-key']);
+  // Loopback is exempt from the key requirement, exactly as the HTTP and
+  // CONNECT gates are.
+  if (!auth.ok && isLoopbackAddr(socket?.remoteAddress)) return { ok: true, client: null };
+  return auth;
 }
 
 /**
--- a/test/upgrade-auth.test.js
+++ b/test/upgrade-auth.test.js
@@ -28,6 +28,27 @@ test('the upgrade gate answers like the other two', () => {
   assert.deepEqual(resolveUpgradeAuth({ headers: {} }, remote, {}), { ok: true, client: null });
 });
 
+test('the loopback exemption is refused to a web page: foreign Origin or Host', () => {
+  // A page can open a WebSocket to 127.0.0.1 with no CORS check, and the
+  // handshake arrives loopback-sourced. Browsers always send Origin on a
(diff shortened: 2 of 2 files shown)


MagicalTux (commented):
Thanks — the gate itself is right and mirrors the CONNECT one, and the wired socket test is exactly what this needed. Two things before it can land, one of them blocking:

**Blocking: the key offered in `Sec-WebSocket-Protocol` is forwarded to the upstream.** `relayUpgrade` copies every header except `host`, the h2 pseudo-headers and (since #311 on master) `x-api-key`. With this PR, a browser that authenticates via `Sec-WebSocket-Protocol: teamclaude, <key>` has that header relayed verbatim, so the proxy's key (or a `clientKeys` entry) is sent to `api.anthropic.com` or whatever `upstream` names — the same class of leak #311 closed for `x-api-key`. Please remove the entry that matched (and drop the header if the list becomes empty) before forwarding, and extend the socket test to assert the upstream sees neither `x-api-key` nor the key in `sec-websocket-protocol`. Returning the matched entry from `resolveUpgradeAuth` is probably the least awkward way to know which one to strip.

**Nit:** the existing JSDoc block for `relayUpgrade` now sits directly above `resolveUpgradeAuth`'s own doc block, so the first comment describes the wrong function. Move it back down above `relayUpgrade`.

**Optional, fine as a follow-up:** a page can open a cross-origin WebSocket to `ws://127.0.0.1:<port>` with no CORS check, and the loopback exemption admits it. Master's request path now refuses key-less loopback callers that carry a browser `Origin`/`Sec-Fetch-Site` (#311); a browser WebSocket handshake always carries `Origin` and CLI clients never do, so the same check would close that door here at no cost.

The branch is behind master (no conflict with this hunk as far as I can see), and branch protection requires an up-to-date branch with green CI before merge, so a rebase is needed anyway.

Frozen666:
An independent review found a real defect in the first version of this, and I have force-pushed a fix. Recording it here rather than quietly amending, because one of the errors is in what I wrote above.

**The `Sec-WebSocket-Protocol` branch leaked the operator's proxy key to the upstream.** `relayUpgrade` strips `x-api-key` but not `sec-websocket-protocol`, so a key accepted from that header was forwarded verbatim to `api.anthropic.com`. Measured on the branch as it stood:

```
subprotocol accepted: true | upstream saw sec-websocket-protocol = "teamclaude, <key>"
x-api-key accepted:   true | upstream saw x-api-key = undefined
```

`test/upstream-credential-hygiene.test.js` exists to pin exactly this invariant, and its own header says it covers "the operator's proxy key". My change routed around that assertion by opening a channel the assertion does not name.

**And my quotation of `relayUpgrade` in the description above was wrong.** I wrote the strip list as:

```js
if (lk.startsWith(':') || lk === 'host') continue;
```

The actual line is:

```js
if (lk.startsWith(':') || lk === 'host' || lk === 'x-api-key') continue;
```

The clause I dropped is precisely the one that makes "not a credential leak" true — and precisely the one the new channel needed to mirror and did not. The severity framing in the description was built on a snippet with the relevant term deleted. I have corrected the description; the same misquote is in #314 and I have corrected it there too.

Three further points from the review, all addressed:

- **The branch had no consumer.** `teamclaude attach` opens no WebSocket at all — `RemoteControl` is `fetch`-only polling against `/teamclaude/status`. My doc comment claimed otherwise. The subprotocol path was speculative support for a browser client that does not exist in this repo, at the cost of a real leak.
- **It widened brute force**, from one guess per connection to roughly four thousand in a single 16 KiB header, where `resolveConnectAuth` tries at most two candidates.
- **A refusal now logs.** A WebSocket client discards the status line, so the 401 alone would have left an operator with a silently dead channel — the same failure shape this gate could cause if some client turns out not to send the key.

The gate is now `x-api-key` only, which is what this issue's own suggested patch proposed. A test pins the subprotocol refusal so the idea cannot quietly return. Mutation-checked: opening the gate, removing the loopback exemption, or dropping the 401 write each turn tests red. 1508/1508 on Node 20.19, 22.19 and 24.12.

Frozen666:
Our messages crossed — your review landed at 14:32, my force-push at 14:38, from an independent review that found the same leak. Thank you for catching it either way; two people finding it separately is the right outcome for a credential path.

I resolved it differently from your suggestion, so please tell me if you would rather have the version you proposed.

**I dropped the `Sec-WebSocket-Protocol` channel entirely rather than sanitizing it.** Stripping the matched entry before forwarding works and is what I would do if the channel had to stay, but three things pointed at deletion instead:

- **It had no consumer.** I wrote in the description that `teamclaude attach` sends the plain header. That was wrong — `src/tui-remote.js` contains no WebSocket at all, it is `fetch` polling. Remote Control sends `x-api-key` directly. So the branch existed for a browser client that does not exist.
- **The offer list is attacker-sized.** `Sec-WebSocket-Protocol` accepts a comma list inside a 16 KiB header, so trying every entry turns one guess per connection into thousands, where `resolveConnectAuth` tries at most two. Sanitizing the leak leaves that widening in place.
- **The proxy cannot honour the negotiation anyway.** It relays the handshake rather than answering it, so any subprotocol it consumed would still have to reach the upstream to be agreed, or be silently dropped from a contract the client thinks it has.

A test now pins the refusal so the idea cannot quietly return. If you want browser support later, sanitize-and-forward is the right shape, and I would rather add it together with the client that needs it.

**I took your test suggestion in the form that still applies.** The socket test now captures the upstream request headers and asserts that neither `x-api-key` nor any other header carries the key onward. Removing the `x-api-key` clause from `relayUpgrade` fails it, alongside `test/upstream-credential-hygiene.test.js`.

**Nit fixed** — `resolveUpgradeAuth` now sits above `relayUpgrade`'s doc block rather than between it and its function.

Two further notes:

- My quotation of `relayUpgrade` in the description omitted `|| lk === 'x-api-key'`, i.e. exactly the clause `be78434` added and exactly the one my change needed to mirror. The severity framing was built on that mutilated snippet. Description and #314 are corrected, and there is a comment above recording it.
- I checked whether the MITM path needs the same gate, since it wires `relayUpgrade` onto its own terminating server. It does not: `srv` is only reachable after `createConnectHandler` passes `resolveConnectAuth`, which answers 407 otherwise. This PR closes the only upgrade entry point reachable without a credential.

Rebased onto `42e8872`. eslint clean, 1514/1514 on Node 20.19.0, 22.19.0 and 24.12.0.

PR #326, opened 2026-09-08
fix(account-pairing): claim a disk row on org evidence, not a partial identity match

lexfrei:
**Please do not merge this until the title says it is ready.** It is up early because the bug it describes is live on master and you should see it now, not because it is finished.

`#323` shipped a regression on the `importFrom` path, and it needs a conjunction rather than any two accounts: one person with entries in two organizations, the earlier entry's organization never stored, and the organization-bearing disk row written ahead of the organization-less one. Disk order is whatever the last writer left, so this is reachable rather than contrived. When it holds, the two entries merge each other's disk row, and `importFrom` names the credentials file an entry reads at the next start. Measured against the file on master:

```
config [ A: uuid U, no org ]  [ B: uuid U, org O2 ]
disk   [ row: org O2, importFrom /second-org ]  [ row: no org, importFrom /legacy ]

on 8bb99d6   ["/second-org", "/legacy"]     A and B hold each other's file
on b46e73d   B kept its own "/second-org"
correct      ["/legacy", "/second-org"]
```

So for this input master is worse than it was before `#323`. That is on me.

`sameIdentity` returns true three ways, not two: an account uuid with both organizations known and equal, an account uuid with the organization missing on either side, and a bare name match. `#323` split off the name case and left the middle one collapsed, so an entry whose organization was never stored matches a row belonging to any organization of that person. Claiming consumes, so the row is then gone from the entry it named.

The fix adds a strict pass ahead of the uuid pass, using `orgKey`, so the ladder is exhaustive by construction rather than by anyone having enumerated the cases correctly. Six tests in the suite plus one new one, mutation-checked: reverting the strict pass reddens the new test alone.

Two things about `#323` worth recording, since both are permanent.

Its commit message says "one entry can no longer inherit another's `importFrom` and read its credential at the next start". The regression above disproves it. A universal claim about a credential path should have rested on what the code guarantees by construction, and that sentence rested on my enumeration being complete. It was not, twice.

It was merged from draft while a review round was still open, and that round is what found this. I published the draft to make the work visible, so the half-finished state was mine to put where it could be merged, not yours to guess about. Hence the shouting in this title: the signal that failed was implicit, so this one is not.

### DIFF
--- a/src/account-pairing.js
+++ b/src/account-pairing.js
@@ -14,7 +14,7 @@
 // evidence and admits no ambiguity: it survives the refresh that rewrites the
 // credential, and it separates two entries that agree on everything else.
 
-import { sameIdentity } from './identity.js';
+import { orgKey, sameIdentity } from './identity.js';
 
 /**
  * The manager account built from config entry `acct`, or null if it has none.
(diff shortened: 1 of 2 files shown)


lexfrei:
@MagicalTux and now it's really ready. The fix is here

PR #344, opened 2026-09-09
feat(docker): add container image and ghcr.io publishing

notmgn:
Adds a Dockerfile so teamclaude can run as a container, plus a workflow to publish images to ghcr.io on version bumps. No source changes.

- Dockerfile: node:22-alpine, tini as PID 1, non-root, healthcheck, OCI labels
- docker-entrypoint.sh: matches the runtime UID to the bind-mounted config's owner, then drops privileges via su-exec
- .dockerignore: keeps config.json and keys out of the build context
- .github/workflows/docker.yml: amd64 + arm64, pushes on version bump with build provenance; PRs build but never push

A plain USER node breaks bind mounts, since the host's ownership overlays the image's and the config comes back EACCES unless chowned to 1000:1000 first. The entrypoint reads the mount's actual ownership and drops to that UID instead. TEAMCLAUDE_UID overrides.

The entrypoint also redirects stderr to stdout so everything lands in docker logs as one stream; TEAMCLAUDE_SPLIT_STDERR opts out.

Usage:

```bash
docker run -d --name teamclaude -p 3456:3456 \
  -v ~/.config/teamclaude.json:/data/teamclaude.json \
  ghcr.io/karpeleslab/teamclaude:latest

docker exec -it teamclaude teamclaude login --token
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,32 @@
+# The image only needs src/, package.json and LICENSE.
+
+# Secrets -- config.json holds account credentials.
+config.json
+*.pem
+*.key
+.env
+.env.*
+
+# VCS and CI
+.git
+.gitignore
+.github
+
+# Dependencies and build noise
+node_modules
+npm-debug.log*
+*.log
+
+# Dev-only tooling and docs
+test
+docs
+screenshots
+nix
+flake.nix
+flake.lock
+eslint.config.js
+.claude
+README.md
+SECURITY.md
+Dockerfile
+.dockerignore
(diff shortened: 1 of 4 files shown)


MagicalTux:
Thanks for this. Reviewed and it looks good. Branch protection here requires a PR to be up to date with master before merging, and this branch does not allow maintainer edits, so I could not update it. I carried your commit unchanged into #354 with a follow-up that pins the workflow's actions by commit SHA (as the other workflows do) and adds docs. Once that merges I'll close this one; the credit stays yours via the kept commit and a co-author line.

MagicalTux:
Merged via #354 with your commit kept intact. Thanks again.

PR #363, opened 2026-09-11
fix(probe): keep Codex tokens off Anthropic usage endpoint

Todmy:
The background OAuth quota prober previously treated every OAuth account without an upstream as an Anthropic target. Codex OAuth accounts also match that shape, so their ChatGPT token was sent to Anthropic /api/oauth/usage and returned HTTP 401.

Gate the target through the account provider and add a regression test asserting Codex accounts are not probed and report not-applicable. Anthropic probing and third-party backend quota probing remain unchanged.


Todmy:
Superseded by clean PR #364, based on current master to avoid conflicts after #361 merged.

PR #364, opened 2026-09-11
fix(probe): keep Codex tokens off Anthropic usage endpoint

Todmy:
The background OAuth quota prober previously treated every OAuth account without an upstream as an Anthropic target. Codex OAuth accounts also match that shape, so their ChatGPT token was sent to Anthropic /api/oauth/usage and returned HTTP 401.

Gate the target through the account provider and add a regression test asserting Codex accounts are not probed and report not-applicable. This branch is based on the current master and also carries the dashboard reload/probe controls and launcher that were not part of the merged dashboard PR.


Todmy:
Superseded by PR #368, which is based on current master and adds the Codex /wham/usage background probe.

PR #378, opened 2026-09-12
fix(threads): refuse thread:continue on upstreams without thread state

lexfrei:
Claude Code keeps the conversation server-side once a message thread exists. First `/v1/messages` body carries `thread: {"type": "create"}` with the whole messages array, every later one carries `thread: {"type": "continue"}` with just the new delta, and the history is looked up by `previous_message_id`.

An upstream that keeps no thread state ignores the field and answers the delta on its own. From the second turn on the model stops seeing the conversation, and nothing errors anywhere. Against a third-party backend I measured input context 23284 tokens on the first turn, then 7596 and stuck there, with the model answering that there is no user message. Threading only kicks in when the client thinks it talks to Anthropic, so this is the MITM path, base-URL routing is fine.

Anthropic answers 400 when a thread can't be continued and the client then resends the whole conversation rather than giving up. This refuses a continue bound for an account with a per-account `upstream` with the same 400. The body carries `details.error_code: thread_unsupported_request`, which the client reads as "this model keeps no thread state": it resends the turn in full and then drops the `thread` field for the rest of the session, so refusals are counted per agent and model instead of per turn, and cost no tokens. `count_tokens` is never refused, there is no conversation to resend for a token count.

An `upstream` whose host is Anthropic's own is a region pin or a mirror and reaches the real thread store, so it is left alone without a flag. Any other relay that does keep thread state opts out with `messageThreads: true`, picked up live on reload.

Not covered: a fleet sent to a third-party host through the global `upstream`, and there is no setting to arm it there. The client-side fallback is undocumented, observed on 2.1.269.


lexfrei:
Running this in real use now. Moonshot and Anthropic accounts share one setup and it works again.

Issue #382, opened 2026-09-13
Prevent teamclaude env from leaking the MITM proxy to unrelated tools


Todmy:
## Problem

`eval "$(teamclaude env)"` exports TeamClaude's MITM listener through shell-wide `HTTP_PROXY`, `HTTPS_PROXY`, `http_proxy`, and `https_proxy` variables. Every unrelated child process then inherits that routing, not only Claude Code.

This breaks developer tools in sandboxed agent shells. For example, `gh api user` is sent to a localhost proxy, where the sandbox denies the private-network connection and returns a synthetic policy 403. The same failure can affect git, package managers, CI helpers, and any HTTP client that honors proxy environment variables.

Adding GitHub to `NO_PROXY` is not a general fix. On machines without a direct route or DNS path to the internet, bypassing TeamClaude fails differently.

## Reproduction

1. Start TeamClaude and run `eval "$(teamclaude env)"`.
2. Run an unrelated HTTP client, for example `gh api user`.
3. Observe that it inherits the TeamClaude localhost proxy.
4. Compare with `teamclaude run`, which scopes the proxy to the Claude child process.

## Expected behavior

The shell setup command should not globally route unrelated tools through the TeamClaude MITM listener.

- `teamclaude env` uses `ANTHROPIC_BASE_URL` only and leaves proxy variables untouched.
- Whole-shell MITM is opt-in, for example `teamclaude env --mitm`.
- Re-evaluating the safe form removes only stale TeamClaude loopback proxy variables from an older shell setup. It preserves a real corporate proxy.
- `teamclaude run` keeps its current process-scoped MITM behavior.

## Validation

A local fork prototype implements this contract in commit `2a01884`:

- 181 tests pass.
- ESLint passes.
- CLI tests cover default mode, explicit MITM, contradictory flags, and stale loopback cleanup.
- The prototype is not pushed to this repository.

This is a general shell environment isolation issue, not a GitHub-specific issue.

PR #384, opened 2026-09-13
fix(env): isolate shell tools from TeamClaude MITM

Todmy:
Fixes the shell-wide proxy leak from `teamclaude env`.

`teamclaude env` now uses base-URL routing by default, so unrelated child tools do not inherit TeamClaude's localhost MITM proxy. Stale loopback exports from an older TeamClaude setup are removed without touching a real corporate proxy. Whole-shell MITM remains explicit with `teamclaude env --mitm`; `teamclaude run` keeps process-scoped MITM behavior.

Updated the shell/proxy documentation and added CLI and environment tests. Related issue: KarpelesLab/teamclaude#382


MagicalTux:
Thanks for this, and for #382, which described the problem precisely. Taken in a different shape in #405, now merged, so closing this one.

The reason not to flip the default: base-URL mode cannot serve the Codex CLI, which honours only proxy variables, nor the hard-coded api.anthropic.com endpoints, and env and run are documented as agreeing on MITM. What #405 does instead is make the default a setting, defaultClientMode: "mitm" | "base-url" (MITM unless set), read by both run and env, with --mitm and --no-mitm deciding per launch and a contradiction refused; the TUI settings screen has a Client mode row for it. Your stale-export cleanup (unset of a proxy variable an earlier MITM eval left pointing at this proxy, leaving a foreign proxy alone) and your subprocess test harness were kept, with a co-author line. An operator who lives in a sandboxed shell sets the mode once and gets exactly the behaviour this PR proposed.

PR #390, opened 2026-09-14
feat(routing): optional one-hop failover on any unclassified refusal

hedgehogcandy:
`forwardRequest` has a branch per upstream status whose meaning is known: 429 is
quota, 403 is the account refused, 401 is a dead credential, 5xx is the provider.
Anything else reaches the client as it arrived.

ChatGPT refuses with one of those others. **"Selected model is at capacity. Please
try a different model."** arrives as a plain 4xx with a `detail` body — verified
against the live endpoint, which answers request-level refusals with a status and
a JSON body rather than an error event inside a 200 stream:

```
POST /backend-api/codex/responses  →  400  {"detail":"Unsupported parameter: ..."}
```

Nothing in this file reads that as a reason to try elsewhere, so the request dies
on whichever account drew it even when a sibling would have served it.

## What changed

`failoverOnAnyError` (default **false**) adds one hop for exactly that case: on any
status >= 400 the branches above did not already claim, pick an alternate the
request has not tried and forward once.

**One hop**, on the reasoning the 5xx path already states — a second account
answering the same way is the provider talking, not the account, and walking the
fleet would spend every account's cache to learn it. The bound is what keeps a
malformed request from becoming one failure per account.

**The account is not marked.** A status this cannot classify is as likely the
request's fault as the account's; a bad body is refused identically everywhere,
and cooling the pool over one such request would take the fleet out for something
no account did wrong. A test pins that the account stays in rotation.

**Default off**, so a pool that has not asked for it sees no change: the
unclassified refusal is returned as it is today, with no sibling spent. A test
pins that too.

## Tests

`npm test` 1792 pass / 0 fail, `npm run lint` and `npm run typecheck` clean,
strict-mode diagnostics on `src/server.js` unchanged at 204.

Reverting `src/` turns two of the four added tests red; the other two are the
off-by-default and no-marking guards, which hold either way by design.

Verified live against a five-account Codex pool with the option on — forcing a
400 through the proxy logs the hop and still surfaces upstream's own status once
the sibling refuses too:

```
[TeamClaude] Upstream 400 on "acct-1" — failing over once to "acct-2" (failoverOnAnyError)
```

Independent of #386, #387, #388 and #389.


hedgehogcandy:
## Field report: this flag caused a retry storm — do not merge as-is

Running this on a five-account Codex pool for a day turned up a failure mode the tests do not cover, and I want it on the record before anyone merges.

**A 4xx that every account shares gets re-sent once per request.** The hop is bounded to one per request, as designed — but a client that retries on its own turns that into a steady doubling. The proxy logged **584 4xx responses in a day**, the large majority on `/v1/messages`, where a malformed or unsupported request is refused identically by every account. Each one spent a sibling to learn what the first refusal already said.

The bound was the wrong shape. One hop per *request* does not bound anything when the requests keep coming; what it needed was to not hop at all on a status that carries no account-specific meaning.

**And it did not catch the case it was written for.** The refusal I built this for — "Selected model is at capacity" — never arrives as a status at all. It comes back **200**, with the failure reported as an SSE event inside the body (measured: 937 × 200 on `/backend-api/codex/responses`, zero 429, zero 5xx, while clients were being refused). #391 handles that case properly and does not depend on this flag's behaviour for statuses.

**Suggestion:** close this in favour of #391, or narrow it to the statuses that actually say something about the account (5xx is already handled; 4xx is not one of them). I would rather withdraw it than have it merged on the strength of a green test suite that never saw a client retry loop.

Happy to close it myself if you prefer — say the word.

hedgehogcandy:
Closing this — the premise is wrong, and measurement since opening it shows the change does harm rather than good.

## The premise was wrong

I opened this having measured that `/backend-api/codex/responses` reports **request-level** errors as an HTTP status with a JSON body:

```
POST /backend-api/codex/responses  →  400  {"detail":"Unsupported parameter: max_output_tokens"}
```

and generalised from that to the capacity case, which I had not reproduced. That generalisation was wrong.

**"Selected model is at capacity" arrives as `HTTP 200` with the error inside the SSE stream** — the third event, after `response.created` and `response.in_progress`:

```
{"type":"error","error":{"code":"server_is_overloaded"}}
```

followed by `response.failed`. The Codex CLI renders that as `codex_error_info: server_overloaded`. Note also that the Codex response carries **no `content-type` header**, so a `text/event-stream` check does not identify it either.

Since this PR gates on `upstreamRes.status >= 400`, it never fires for the case it was written for.

## The measured side effect

What it does fire on is genuine 4xx — which are the request's fault, identical on every account. Over one day on a five-account pool that produced **323 pointless hops**: a second account spent per malformed request, and its answer discarded, to learn what the first one already reported.

## What the right fix looks like

Peek the Codex response stream before `res.writeHead` (bounded — 256KB was enough), up to the first complete event, and treat `server_is_overloaded` as the retry signal. Re-forward with a different model rather than a different account, since capacity is a property of the model, not the subscription — walking accounts does not help. That approach is running locally here and fires in practice.

Not opening that as a PR right now, since it needs the streaming path reworked (buffering the whole body to inspect it costs ~9s to first byte) and that is a larger change than this one. Leaving the measurement here so the next attempt starts from the right wire format.

#386, #387, #388 and #389 are unaffected — each of those rests on a header or payload reading measured directly, not inferred.

PR #391, opened 2026-09-14
feat(routing): fail over when a 200 stream reports its own failure

hedgehogcandy:
> **Stacked on #390**, whose `failoverOnAnyError` flag this shares. Merge that first and
> this rebases to a single commit; its second commit alone is this change.

Every failover in `forwardRequest` keys on the upstream status, and the Responses API does
not always use one. It answers **200**, opens the SSE stream, and then reports the refusal
as an event in the body — "Selected model is at capacity" arrives exactly that way.

Measured on a five-account Codex pool, from the proxy's own activity log:

```
POST /backend-api/codex/responses   200 × 937   499 × 25   400 × 2
                                    429 × 0     5xx × 0
```

Not one status-shaped failure, while clients were being refused. So the proxy counted
those as answers, no hop was considered, and the refusal reached the client as though it
were the reply — on an account that may have been the only one refusing.

## What changed

Read the head of a streaming response **before its headers go out**. Nothing has been
written to the client at that point, so the request is still retryable.

While the stream is uncommitted (`response.created`, `.queued`, `.in_progress`) keep
reading; the first event that is not one of those decides it. A `response.failed` or
`error` there is a refusal with no output behind it → hop once to a sibling. Anything else
releases the stream untouched.

**Order decides this, not presence.** A failure that arrives after real output is a stream
that broke after committing, and committed output has no retry behind it — that case is
relayed, and a test pins it.

The hold is bounded twice, by a byte budget and by a wall clock, so a slow first token is
never held hostage and a stream that says nothing is released rather than waited on. On
release the bytes already read are replayed ahead of the rest of the same stream, so the
client sees an unbroken body.

One hop, for the reason the status-keyed paths take one: a second account answering the
same way is the provider talking, not the account.

## Tests

`npm test` 1796 pass / 0 fail, `npm run lint` and `npm run typecheck` clean, strict-mode
diagnostics on `src/server.js` unchanged at 204.

Reverting `src/server.js` turns two of the four added tests red; the other two are the
already-committed and off-by-default guards, which hold either way by design.

Verified live against the same pool — the normal path is unchanged and shows no added
latency to first byte.


MagicalTux:
This branch still carries `ca5ef220 feat(routing): optional one-hop failover on any unclassified refusal`, the commit from #390, which you closed as doing harm rather than good, and the `failoverOnAnyError` flag it introduced appears 15 times in this diff. Could you rebase this onto master without it? The stream-failure change itself reads well and I would like to take it on its own.

PR #393, opened 2026-09-14
feat(auth): let a client run against the pool without a login of its own

hedgehogcandy:
The proxy rotates which account serves a request, but the client still has to
authenticate as somebody: every machine running `teamclaude run` needs its own
Claude login. On a shared proxy that is the wrong shape. The accounts are
already in the pool — a second machine should be able to spend them without a
human signing into it, and without that machine ever holding a token it could
take elsewhere.

`proxy.managedClientAuth` (off by default) makes `run` and `env` hand the CLI a
facade bearer instead:

```
tc-managed-v1.<base64url account id>.<HMAC-SHA256 over proxy.apiKey>
```

It names an identity and carries nothing else — no access token, no refresh
token, no OAuth origin. The proxy verifies the signature, resolves the identity
to a live account, and injects the real token upstream, so the credential is
inert anywhere but the proxy that minted it.

### Revocation

Disabling the mode, removing or disabling the account, or rotating
`proxy.apiKey` invalidates every issued credential at the next
`POST /teamclaude/reload`. That is why identity metadata is re-read from disk
there rather than inferred from the account merge, which is deliberately
additive — only eligibility metadata is kept, never tokens or URLs.

### Three refusals

Each closes a way the facade could become a real credential, or leak one:

- **`--no-mitm` is refused, not downgraded.** Base-URL routing leaves the CLI
  able to reach `api.anthropic.com` directly, where the bearer is not a
  credential; the failure would read as a broken login rather than a
  misconfigured mode.
- **Anthropic OAuth accounts only** are eligible identities. A facade bearer
  aimed at a non-Anthropic upstream — including over a WebSocket upgrade, which
  MITM also serves — is refused 403 instead of relayed.
- **The 23 environment variables** that let the CLI choose another credential or
  provider (`ANTHROPIC_API_KEY`, `ANTHROPIC_BASE_URL`, `CLAUDE_CODE_USE_BEDROCK`,
  `CLAUDE_CODE_OAUTH_REFRESH_TOKEN`, …) are unset **in the child only**, and
  `NO_PROXY` entries matching an intercepted host are dropped: bypassing the
  proxy would strand the CLI holding a token nothing else accepts. The parent
  environment and the user's saved login are copied, never mutated.

Signature comparison is `timingSafeEqual`, and the account id must survive a
base64url round-trip, so no two encodings can name one identity.

### Tests

`test/managed-client-{auth,proxy,reload,run}.test.js` — credential issue/verify
including malformed, revoked and non-canonical forms; the proxy and MITM paths;
live disk reload revoking a token while normal inference continues; and the
`run`/`env` launch environment, asserting no provider credential is written and
that `--no-mitm` is refused. Full suite green, plus `npm run lint` and
`npm run typecheck`.

Docs: `proxy.managedClientAuth` in `docs/configuration.md`, and a
"Pool-managed authentication" section in `docs/usage.md`.


MagicalTux:
I read this closely with the maintainer. The mechanism is coherent: the facade carries one signed account id, every identity-plane request (`/api/oauth/*`, `/v1/code/*`, the Remote Control WebSocket) is served with that one account's token and never rotated, inference still rotates, and revocation is live on reload. That stability is exactly the property Remote Control needs, so the design is right.

Two things are needed before it can go in, plus a rebase (the branch predates several merges and its CI run is red):

1. **A deliberate choice of the identity account.** `issueManagedCredential` takes the first eligible account unless `TC_ACCT` names one, so reordering the config silently changes which identity Remote Control runs under. Please add a config key, for example `proxy.managedIdentity` naming an account, refuse to issue when it is set and not eligible rather than falling back, and show the chosen identity in the launch line and in the `managedClientAuth` status block.
2. **The failure mode made visible.** When the identity account is in error or its token is rejected, identity-plane calls answer 503 and the CLI most likely does not start. Status and the dashboard should say "managed identity `x` unavailable: re-login" distinctly from "proxy down", and `teamclaude run` should check this before launching and print that line instead of leaving the user with a CLI that fails at startup.

It also rests on Claude Code accepting an arbitrary string in `CLAUDE_CODE_OAUTH_TOKEN`; a sentence in the docs saying that this is observed behaviour of a given CLI version, not a documented contract, would be honest. This would close #395.

PR #410, opened 2026-09-16
fix(tui): a dead terminal must not take the proxy with it

rikbrown:
Follow-up to #404, which I authored: that change fixed the stalls but introduced a crash. This fixes it.

## What #404 did, and what it cost

Flipping stdout non-blocking stopped a full-screen paint from holding the event loop. It also moved stdout's write failures onto the **asynchronous** path, where a failure arrives as an `'error'` event on the stream rather than throwing at the call site. Nothing listened for it, so Node promotes it to an `uncaughtException`, `installCrashHandlers` records it and calls `exit(1)`.

The result: **the proxy dies whenever its terminal goes away** — a pane closed, a pty recreated, a terminal emulator restarted. Worse, the hard exit skips `stop()`, so a supervised sidecar is orphaned and keeps its port, and the replacement server then cannot bind (`Address already in use`) and restart-loops indefinitely.

## Evidence

Two crashes within a day on a live deployment of my fork, both identical:

```
=== 2026-09-15T21:58:50.452Z uncaughtException ===
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)
```

`WriteWrap.onWriteComplete` is the async write-completion path, which only exists once the stream is non-blocking. The crash log (added in #150) had been armed for **six weeks with no entries**; its first is thirteen minutes after #404 first ran.

## The fix

A display may no more kill the proxy than block it.

- `start()` installs an `'error'` listener before the first write and records that stdout is broken.
- `_paint()` returns early when it is, so the process serves on without a display — and a stream that never drains cannot strand the pending-paint handshake.
- `stop()` makes the exit sequence best-effort, since a terminal that has already gone will fail it, and releases the listener *after* that write so a late async failure is still absorbed.

`index.js` already carried this idiom for the status command's EPIPE; the TUI never needed it until stdout stopped blocking.

## Verification

- `npm test` — 1868 pass, 0 fail (3 new regression tests).
- `npm run typecheck` — clean.
- `npm run typecheck:strict -- --base upstream/master` — `1831 strict-mode diagnostics (base 7b1e92d0: 1831)`, no regression.
- `npm run lint` — clean.

The new tests drive the real failure: emit `'error'` on stdout and assert that painting stops and nothing throws, including when a frame is already parked awaiting `drain`.


rikbrown:
Pushed a follow-up commit: **the patch as originally submitted had a hole, and I'd rather you saw the corrected version than the one that bit me.**

`stop()` removed the stdout `'error'` listener immediately after writing the exit sequence, with a comment asserting that the listener outlived that write. It does not — the removal is synchronous and the failure is not. Flipping back to blocking also does not make writes *already queued* synchronous, and shutdown continues well past that point (background workers, then an awaited state save). A paint still in flight fails into that window with nothing listening.

That killed the proxy a third time, on the deployment I took the original evidence from:

```
=== 2026-09-16T15:19:12.435Z uncaughtException ===
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)
```

Two changes:

- The listener now lives as long as the process. It only sets a flag, and `start()` is called once, so nothing accumulates.
- **`crash-log.js` no longer treats an uncaught `EPIPE` as fatal.** This is the fix that generalises: a dead terminal says nothing about the proxy's state, so patching teardown orderings one at a time is chasing the symptom. It is recorded once — with `err.code`/`err.syscall`, since a bare "Error: write EPIPE" names neither the stream nor the call — and then survived. A genuine uncaught exception still reports and exits non-zero. `reportFailure` in `server.js` already documents this hazard ("this daemon treats an uncaught EPIPE as fatal"); this makes that no longer true.

If you would rather keep `crash-log.js` strictly faithful to Node's exit-on-uncaught behaviour, say so and I will confine the change to the TUI listener — but then a headless server writing to a closed stdout still dies, so I think the broader fix is the right one.

Gate on this branch: `npm test` 1872 pass / 0 fail, `typecheck` clean, `typecheck:strict --base upstream/master` `1831` vs base `1831`, `lint` clean.

PR #411, opened 2026-09-16
fix(server): outlive the client's connection pool on idle keep-alive

rikbrown:
## The race

Node closes an idle keep-alive connection after **5s** (`server.keepAliveTimeout`). A client whose own connection pool holds that socket for longer will eventually write to a connection the server has already closed, and the request fails before it reaches any upstream. Whoever closes first wins; when it is the server, the client finds out only by failing.

Setting the server's idle timeout above the longest client pool makes the client always the side that closes, which is the side that can do it safely.

## Evidence, and its honest limits

My evidence comes from a deployment of my fork whose client is **reqwest**, whose `pool_idle_timeout` defaults to 90s and which never overrides it. Two requests in the error corpus failed at **126ms** and **161ms** with `error sending request` and no upstream involvement at all — the signature of writing to a dead socket.

That particular client is fork-only, so I am not claiming this is biting upstream users today: Claude Code's undici pool defaults *below* 5s and will rarely lose the race. But the exposure is general — any client with a pool longer than 5s (a Rust or Go service, curl with connection reuse, a proxy in front of this one) hits it, and the failure is silent, fast and looks like a network fault rather than a timeout. The cost of the fix is one idle socket held per idle client.

Happy to drop this if you would rather upstream kept Node's default and this stayed a fork concern.

## Why `headersTimeout` is untouched

It would be natural to assume `headersTimeout` (60s) also bounds the idle gap between keep-alive requests, and that raising one requires raising the other. I measured it rather than assume: with `keepAliveTimeout` 3s and `headersTimeout` 1.5s, an idle socket still survived to ~4s. So `headersTimeout` bounds an in-progress request's headers only, and raising it would weaken a slowloris bound for no benefit.

## Verification

- `npm test` — 1867 pass, 0 fail (2 new).
- `npm run typecheck` — clean.
- `npm run typecheck:strict -- --base upstream/master` — `1831 strict-mode diagnostics (base 7b1e92d0: 1831)`, no regression.
- `npm run lint` — clean.


rikbrown:
Pushed a second commit: this patch fixed only half the proxy.

The 120s hold went on the base listener created in `server.js`. The MITM path terminates a server of its own (`http2.createSecureServer({ allowHTTP1: true })` in `mitm.js`) and that one kept whatever the runtime defaults to — so one proxy answered to two different numbers, and only one of them was deliberate. It is also the half that matters most in practice: every session started by `teamclaude run` reaches the proxy through a CONNECT tunnel, not the base URL.

Measured, rather than reasoned about:

| | base listener | MITM server |
|---|---|---|
| node 24.12.0 | `Keep-Alive: timeout=120` | no header; idle connection never closed |
| node 26.8.1 | `Keep-Alive: timeout=120` | `timeout=5`; idle connection closed at 6.0s |

So the exposure varies by Node version, which is exactly the kind of thing worth pinning with a test rather than a constant. The new test drives a real CONNECT + TLS + HTTP/1.1 exchange through the tunnel and asserts the `Keep-Alive` header the client's pool actually schedules against; it fails without the fix.

One wart worth flagging for review: `@types/node` does not declare `keepAliveTimeout` on `Http2SecureServer`, although the runtime honours it for the HTTP/1 connections `allowHTTP1` accepts. I used a cast and leaned on the end-to-end test to keep it honest. If you would rather see it set some other way, say so and I will change it.

Issue #407, opened 2026-09-16
Synthetic 429 claims all accounts at quota when one is credential-error with remaining Fable headroom


jpeslar1:
## Summary

When rotating multiple Claude Max OAuth accounts, the synthetic 429 from `exhaustedMessage` still says every account is "at their quota or rate limit" even when one account is in `status=error` (dead refresh token / `invalid_grant`) and still shows remaining Fable weekly headroom in the TUI.

This is extremely misleading in day-to-day operation: operators wait for a quota reset (or retry for 60s) while the real fix is `teamclaude login` / `teamclaude import`.

Related: #168 / #270 improved the exhausted wording for disabled accounts + model naming, but still conflate auth failure with quota.

## Observed client error

```
API Error: Server is temporarily limiting requests (not your usage limit) ·
No account can serve this request for claude-fable-5-1: all 2 accounts are at
their quota or rate limit. Quota resets in 60s.
```

While this was showing, the live fleet looked like:

| Account | Plan | Fable weekly | Status |
|---|---|---|---|
| john@zevari.ai | Max 5x | ✗ ~100% used | active (genuinely spent) |
| jpeg340@gmail.com | Max 20x | ✓ ~12–13% used | **error** (`invalid_grant` / revoked refresh token) |

So one account had real Fable capacity left, but was skipped because credentials were dead. The synthetic 429 blamed quota for both.

## Why this happens repeatedly

1. Claude Code `/login` (or another tool holding the same OAuth family) rotates the refresh token.
2. teamclaude still holds the previous refresh token → Anthropic returns `invalid_grant`.
3. Account is marked `status=error` and correctly skipped by selection.
4. Quota bars for that account still render (Fable looks available).
5. The other account may be genuinely spent on the Fable weekly bucket.
6. Client gets a synthetic 429 whose text says **all** accounts are at quota.

Operators then chase the wrong problem for minutes (retry loop attempt 1/10, "wait 60s") instead of re-auth.

## Expected

The synthetic 429 should:

- Name credential-error accounts separately and point at `teamclaude login` / `teamclaude import`
- Name quota/rate-limited accounts separately
- Not claim a quota reset will clear an auth failure
- Not say "all N accounts are at their quota" when some are auth-dead with remaining headroom

## Environment

- teamclaude via LaunchAgent / `teamclaude run`
- 2 OAuth Max accounts rotating for `claude-fable-5-1`
- macOS

## PR

[...]

PR #417, opened 2026-09-18
fix(oauth): sync Claude Code refresh tokens onto the fleet

jpeslar1:
## Summary

- Claude Code refreshes OAuth against `platform.claude.com`, which TeamClaude previously **blind-tunneled**. Those refreshes rotate the refresh-token family while TeamClaude still held a copy → `invalid_grant` → account dropped from rotation even though Claude Code / Keychain still had a working login.
- MITM now terminates `platform.claude.com` in **`oauth-mirror`** mode: passthrough (no fleet token inject) + sync successful `refresh_token` grants onto matching fleet accounts.
- The same sync runs when `/v1/oauth/token` hits the inference upstream.
- On `invalid_grant`, re-read the Claude Code store (`importFrom` / macOS Keychain) before marking the account as needing re-login.
- Docs: prefer `importFrom: "~/.claude/.credentials.json"` for any account Claude Code is also signed into.

No personal account emails or tokens in this PR.

## Test plan

- [x] `node --test test/oauth-sync.test.js test/mitm-provider-hosts.test.js`
- [x] oauth-mirror host mode for `platform.claude.com` (not rewrite)
- [x] sync updates the account that held the sent refresh token
- [x] invalid_grant recovers when Claude Code store has a fresher refresh token
- [x] invalid_grant still errors when the store still holds the dead token
- [ ] Live: Claude Code under `teamclaude run` refreshes → log shows `Synced client OAuth refresh onto account ...` and the account stays active

Made with [Cursor](https://cursor.com)


MagicalTux:
Thank you for the work here, and for #407/#408, which describe a real operator problem well. I am closing this one rather than merging it; the reasons are structural, so reshaping it on the branch would not get it there.

**What it does that the project should not do**

1. **It terminates TLS on the OAuth host.** `platform.claude.com` has been a byte-exact blind tunnel on purpose: the client's login and token refresh "pass through untouched" is a documented contract. In the new `oauth-mirror` mode every non-token path goes through `relayRaw`, which forwards only three headers and buffers the body as text, so the client's authorization-code exchange and its PKCE verifier are parsed in the proxy's memory. That is a much larger trust ask than routing inference.
2. **A fleet account can silently adopt the operator's local login.** `_recoverFromClaudeCodeStore` defaults to `~/.claude/.credentials.json` (or the Keychain) for every Anthropic OAuth account, including ones with no `importFrom`, and does no identity check. Any fleet seat that hits `invalid_grant` therefore takes whatever login Claude Code has on that machine. The entry now names one account and authenticates as another, and the two holders of that token family kill each other on every refresh. `sameIdentity` exists for exactly this comparison and is not used.
3. **It only works in one direction.** When the fleet refreshes first, Claude Code's copy still dies; nothing writes back, and `ownsInlineToken` deliberately skips persisting tokens for `importFrom` entries.

**The premise**

The failure in #407 comes from one OAuth token family being held by two parties: Claude Code's own login and a fleet entry imported from it. Whoever refreshes last wins. The robust answer is a separate `teamclaude login` per fleet seat, so the fleet's tokens and the client's login never share a family; the docs edit in this PR steers users toward the shared setup instead, which I would rather warn against.

**What I would take**

- #408 (the message fix) is being merged, rebased onto #429.
- A narrow recovery: on `invalid_grant`, for an account that explicitly has `importFrom`, re-read that file and adopt its tokens only when `sameIdentity` confirms it is the same account and organisation. No TLS interception, no default path. That is perhaps forty lines and I would merge it gladly.
- A docs paragraph in `docs/accounts.md` saying that an imported login shares a refresh-token family with Claude Code and recommending a separate `teamclaude login` for a long-lived fleet seat.

PR #427, opened 2026-09-18
feat(routing): fall back to extra-usage accounts when the fleet is spent

RuslanTar:
Once every account is past its quota, selection ends in the throttled revalidation probe or in nothing at all, and the client gets a synthetic 429 (or waits out `holdSeconds`). An account with Anthropic's "extra usage" (paid overage) switched on does not stop at its plan limit — so a fleet that has one was refusing requests it could have served.

`accounts[].allowExtraUsage: true` opts an account in:

```json
{
  "name": "team@example.com",
  "allowExtraUsage": true,
  "maxUsage": { "unified7d": 1.5 }
}
```

**This spends real money**, so the whole design is "last resort, and only what was asked for":

- **Rotation is unchanged.** An opted-in account still rotates away at `switchThreshold` like any other, and is never chosen while any account can serve under the normal rules.
- **Used only where the walk would otherwise have nothing** — after the normal walk, and after the free revalidation probe, which still goes first when it is due because headroom it finds costs nothing. A probe upstream refuses on quota re-enters the walk with the probe slot spent, so that same request lands on the fallback instead of 429ing. `pickAlternate` reaches it too, so the 429/5xx failover hops stop ending in a 429 while a paid account sits idle — but its gate is fleet-wide, not the request's `exclude` set: a hop off an account that is merely rate-limited for a minute waits that out as before rather than billing to skip the wait. `pickAlternate` stays side-effect free (#286).
- **Only the soft refusals are overridden** — `quota` (the switch threshold, a spent family bucket included) and a remembered `upstream-rejected` verdict. `disabled`, `maxUsage`, a live 429 hold, an entitlement cooldown, `error`, routes/ownership and the provider partition all still bind, and an account whose usage probe reports overage switched off upstream (`spend.enabled === false`) is skipped since it would only 429. No spend data means upstream decides.
- **Several opted-in accounts** are ordered as the probe orders them: `priority` first, then least utilized.
- **It returns on its own.** An account past its quota is never available, so the first request that finds normal headroom switches back — no new selection code.
- Utilization passes 100% in overage, so `maxUsage` above 1.0 doubles as a spend limit.

Visibility, since paid traffic that looks like idle traffic is the failure mode here:

- Entering and leaving extra usage is logged once per episode, scoped per model and provider (`_cursorKey`), so alternating Fable and Opus traffic does not produce a log line per request or restart the ramp. A failover hop onto a paid account is logged once per episode as well.
- `teamclaude status`: `extra usage allowed` in the account header, and `— serving on extra usage (paid overage), billing` on the `Blocked` line while it is serving.
- TUI (in-process and attach): a yellow `xu` tag, red `xu!` while billing, budgeted into the row width like the existing `$`/`$!` tag.
- Web dashboard: an `extra usage allowed` / `on extra usage — billing` badge.
- `getStatus` carries `allowExtraUsage` and `onExtraUsage` per account.
- The cursor is never left stranded on a paid account: once a normal account can serve again, `getActiveAccount` re-seats it. Without this, a fleet running `distributeSessions` (whose session walks deliberately do not move the shared cursor) kept reporting a paid account as current long after traffic had moved off it.

Config edits apply live on reload through `sync-accounts.js`, so turning the opt-in off stops the spending at once.

Tests: `test/extra-usage-fallback.test.js`, covering the fallback path and every gate that must still refuse, the per-model episode scoping, the failover hops, and the cursor case under `distributeSessions: true` and `"adaptive"`. Full suite, lint, typecheck and the strict-typecheck ratchet all pass.


MagicalTux:
Good idea and carefully built: opt-in only as a literal `true`, hard gates (disabled, capped, entitlement, throttled, error, route) still block, and the docs are loud about cost. I would like to merge it, but one thing has to change first because it spends money.

**Blocking: it can start billing a paid account while a non-billing account still has free quota.** The gate in `_pickExtraUsage` refuses only while another account is *available*, which means under `switchThreshold`. An account between its threshold and 100% still has free quota, and the revalidation probe gives it one request a minute. Your own first test has `a` at 0.99 with no opt-in and selects `paid`. With a per-bucket threshold such as `unified7d: 0.85` the fallback runs a paid account into overage while a free one holds about 15% unused. The README and quota docs say "once every account is spent", which is not what the code does.

What I would like to see:

1. In `_pickExtraUsage`, first return the best soft-refused account (priority, then least utilised, opted in or not) whose maximum utilisation is below 1.0 and which is not upstream-rejected. Pick a paid account only when none is left.
2. Set `onExtraUsage`, the red `xu!` tag and the "billing" wording only at utilisation of 1.0 or more, or when `spend.usedMinor` is rising. Today they show for a spent-but-not-billing account.
3. Expire a billing episode when the paid account becomes available again, not only on a later request in the same scope; and key episodes by account id rather than index, since removing an account shifts indices.
4. A test with a normal account at 0.99 asserting the paid account is NOT used, and docs wording that matches (including that an account in `error` or entitlement cooldown counts as unable to serve).

#428 is being merged first and touches many of the same places (`makeAccount`, `getStatus`, both `sync-accounts.js` hunks, `accountBadges`, the TUI row tags, the configuration table), so this will need a rebase afterwards; `unavailableReason` picks up per-account thresholds automatically, so there is no behavioural conflict.

RuslanTar:
Rebased onto master after #428 (conflicts were all keep-both, plus porting the `xu` width reservation into the new `_listLayout`), and the four points are in `057b685`:

1. **Free quota first.** `_pickFallback` now picks in two tiers: the best soft-refused account (priority, then most free quota left, opted in or not) with max utilisation < 1.0 and no upstream `rejected` verdict; a paid account only when none is left. The fleet-wide gate and the "no opt-in → still 429" behaviour are unchanged.
2. **Marked only while billing.** `onExtraUsage` (and so `xu!`, the badge and the status line) is true only for a paid-tier episode whose account is at ≥ 1.0 on a governing bucket, or whose `spend.usedMinor` has risen since the episode began. The free tier is never marked.
3. **Episodes end on their own and follow the account.** A sweep runs from every status read as well as every selection and ends an episode as soon as its account, or any other in the partition, can serve the scope's model normally again. Entries hold the account object rather than an index — `id` is `acct.id || null` on manager accounts, so a reference is the stable handle; a removed account simply stops matching and its episode ends.
4. **Tests + docs.** `a normal account at 0.99 is used before the paid account, which is NOT billed` (and a per-bucket `unified7d: 0.85` variant), billing starting only at 100%, an error/entitlement account counting as unable to serve, episode expiry without a request, and index-shift on removal. README, `docs/quota.md` and `docs/configuration.md` now say what the code does.

2235/2235, lint and typecheck clean, strict ratchet unchanged.

One thing I left as is and would like your view on: while the free tier is serving, `teamclaude status` still shows that account as `Blocked local switch threshold reached` (the cursor names it as current, so it is not invisible). If you want a line for it — e.g. "serving past its threshold on free quota" — I can add an `onFreeFallback`-style flag, but did not want to widen the payload without asking.

Issue #288, opened 2026-09-07
A request-scoped 429 is amplified into a fleet-wide stall: fabricated 60s retry-after plus an account-wide pause


Frozen666:
One client pinned to a model id upstream refuses (`claude-sonnet-4-6` — 429 on every account, any load, any payload) degraded a six-account fleet's other four developers from **p50 4.3s to p50 64.5s for three hours**. The model id is the user's problem. Turning it into everyone's problem is the proxy's, and it happens in three steps that compound.

## The chain

**1. `retryAfter` is fabricated when the header is absent** — `server.js:1678`

```js
let retryAfter = parseInt(upstreamRes.headers.get('retry-after'), 10);
if (Number.isNaN(retryAfter)) retryAfter = 60;
```

The tell that this is what fired: **all 103 logged absorbs read exactly `waiting 60s`, with zero variance.** A genuine per-minute throttle returns scattered `retry-after` values. 103 identical samples means the header was absent every time and 60 is the proxy's own constant. A duration histogram of the episode shows the 58–70s band at 8.2% of requests versus 0.7% at baseline — a 12× enrichment — plus a secondary bump at 115–125s where two absorbs chained.

**2. The whole account is paused for those 60 seconds** — `pauseAccount`, called from the 429 branch. Every *unrelated* request routed to that account then waits in `admit()`, and the ramp re-serializes the backlog for 30s after release. The account is not out of quota and not throttled by upstream; one request was refused.

**3. The doomed request is retried up to `maxRetries` (= account count)**, so 5×60s + overhead ≈ 310s — which matches the observed 329–338s of the wedged requests, after which the client gives up.

## Reproduced in two requests

On an idle fleet at p50 4s:

```
07:44:58  probe: max_tokens:1, claude-sonnet-4-6, pinned to a completely idle account
          → Rate-limit 429 on "<idle-account>" — failing over once to idle account "<other>"
07:44:59  probe: claude-sonnet-5 (known good), pinned to that SAME account
          → hung 45s until aborted, blocked in admit() behind the pause the first probe armed
```

A single bad request makes a healthy account unusable for everything else for a minute.

## The diagnostic message points the wrong way

```
[TeamClaude] Second account rate-limited too — the limit looks IP-scoped, not per-account
```

[...]

PR #431, opened 2026-09-19
fix(server): retry headerless 429 once for callers that wait

rikbrown:
## The gap

A 429 with no `retry-after` or `anthropic-ratelimit-*` headers is request-scoped (#288): no account is paused, the request gets one hop to an idle sibling, and a second such 429 returns to the client without a fabricated `retry-after`.

That treats every refusal as permanent. Most observed refusals were not.

| | count |
|---|---|
| Fable requests drawing a headerless 429 | 4 in 238 (about one every 8 min) |
| Codex requests drawing one | 0 in 164 |
| All other requests drawing one | 0 in 728 |

Over 32 minutes, consecutive refusals arrived 0.6–0.8s apart, from four different starting accounts. The client's own retry usually succeeded after its backoff. Claude Code showed `Waiting for API response · will retry in 2m 38s`, then retried without writing to the session transcript.

## What changed

**One retry after 2s, on the account the hop landed on.** `TEAMCLAUDE_HEADERLESS_429_RETRY_DELAY_MS` is a choice, not a measured limit: it exceeds the observed 0.6–0.8s hop interval and is below the 2m 38s client backoff. The retry:

- runs once per request, with no ladder;
- stays on the current account rather than selecting a third, sets `ctx.hopTo` to preserve the fleet cursor (#286), and retains `route`;
- yields to an armed sx fresh-IP retry, and skips when the reply is on the wire, the client has gone, or the abort signal ends the wait.

If this retry is refused, the existing line still logs and the client receives the 429 about 2s later. The existing no-sibling retry also used a hardcoded 2s; both now use one constant and, as below, one flag.

**One wait per request.** The no-sibling and post-hop retries can both run in one request: no idle sibling causes the first retry, then a sibling can return before that retry is refused. Separate flags let that request wait twice. They now share `ctx.requestScopedRetried`; a test drives this sequence.

**Do not hold a caller that will not wait.** Codex gives the response head a fixed 60s, then retries the request itself. `src/provider.js` now provides `holdsConnection` beside `rewritesBody`: true for Anthropic, false for Codex, and true for an unknown or absent provider to preserve existing behaviour. Both retries check it. The log identifies the reason:

```
429 followed the request onto "b" with no rate-limit headers — codex caller does not wait; returning it to the client
```

This also removes the existing 2s no-sibling wait for Codex callers. No headerless Codex 429 appeared in the observed 164 requests, but such a 429 does classify as request-scoped because the classifier checks `anthropic-ratelimit-*` and `x-codex-*` only.

## Verification

`npm test` (1872 tests), `npm run lint`, `npm run typecheck`, and `node scripts/typecheck-strict.mjs` pass. No test calls a real endpoint.

New coverage in `test/request-scoped-429.test.js` and `test/provider.test.js` verifies:

- a headerless 429 surviving the hop retries exactly once on the hop account (`['t-a', 't-b', 't-b']`), with no third account and an unchanged fleet cursor;
- a retry that clears does not reach the client and waits `TEAMCLAUDE_HEADERLESS_429_RETRY_DELAY_MS`;
- a request reaching both sites waits once, not twice;
- aborts and replies already on the wire skip the retry;
- headerless Codex 429s return immediately both without a sibling and after a hop; and
- `holdsConnection`, including that `'constructor'` is not a provider.

The existing control remains unchanged: a 429 with `retry-after` still throttles and pauses the account.

## Deliberate limitation

`RATE_LIMIT_ABSORB_MAX_SECONDS` can still hold a Codex caller for up to 60s on a 429 with `retry-after`. That separate existing path is not changed here.


rikbrown:
Heads-up: this overlaps #429 in the 429 branch of `src/server.js` (both touch the same block; the merge is a content conflict, not a semantic one). Whichever lands first, I'll rebase the other straight away.

